PR #29, opened 2024-04-16
Start moving chunking into parser land.

oxaroky02:
We're moving chunking to the parsers.

- Implemented a `Parsers::BasicTextChunker` module that provides `chunk_bytes`
- Note the passing of `profile` as parameter rather than initializing parser with profile
- `Chonker` no longer does any chunking
- `Chonker` expects parsers to support `chunk_by_xxx` method for supported chunking methods `xxx`
- `Pdf` and `Text` parsers include `BasicTextChunker`, preserving the prior chunking approach. 
- Using chunking profile from within document instead of passing it to parser

### DIFF
--- a/app/services/chonker.rb
+++ b/app/services/chonker.rb
@@ -1,8 +1,6 @@
 class Chonker
-  # implement basic text chunker until all parsers will be self-chunking
-  include Parsers::BasicTextChunker
-
   class UnknownChunkingMethod < StandardError; end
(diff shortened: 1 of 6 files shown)


oxaronick:
Is the document profile being used still, or is this letting the parser override that?

oxaronick:
> Is the document profile being used still, or is this letting the parser override that?

Ah, nm. I see it.

PR #30, opened 2024-04-23
Change API to allow ingest by iterating over chunks

oxaroky02:
The goal is to iterate over chunks and embed them.

- `Chonker#each` can take a block to yield chunks (or returns iterator when called without a block)
- `TheIngestor#ingest` modified to iterate through chunks and embed them.

Changes to parsers to actually incrementally yield chunks per parser will come afterwards.

### DIFF
--- a/app/services/chonker.rb
+++ b/app/services/chonker.rb
@@ -13,6 +13,17 @@ class Chonker
     @parser = parser
   end
 
+  # Yield each chunk
+  def each(&block)
+    if block_given?
+      chunks.each(&block)
+    else
+      chunks.each
+    end
+  end
+
+  private
+
   def chunks
     raise UnknownChunkingMethod, "Unknown chunking method '#{chunking_method}'" unless CHUNKING_METHODS[chunking_method]
 
@@ -23,8 +34,6 @@ class Chonker
               "Parser (#{@parser.class.name}) doesn't support chunking method '#{chunking_method}'"
     end
   end
-
-  private
 
   def chunking_method
     @parser.chunking_profile.method
--- a/app/services/chonker.rb
+++ b/app/services/chonker.rb
@@ -13,7 +13,6 @@ class Chonker
     @parser = parser
   end
 
-  # Yield each chunk
   def each(&block)
     if block_given?
       chunks.each(&block)
--- a/app/services/chonker.rb
+++ b/app/services/chonker.rb
@@ -2,6 +2,8 @@ class Chonker
   class UnknownChunkingMethod < StandardError; end
   class UnsupportedChunkingMethod < StandardError; end
 
+  include Enumerable
+
   CHUNKING_METHODS = {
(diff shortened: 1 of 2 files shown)


app/services/chonker.rb
```diff
@@ -13,6 +13,17 @@ def initialize(parser)
     @parser = parser
   end
 
+  # Yield each chunk
```
nickthecook, line 16: Do you think this comment is necessary?

app/services/chonker.rb
```diff
@@ -24,8 +35,6 @@ def chunks
     end
   end
 
-  private
```
nickthecook, line 27: Did you intend to make `chunking_method` public?

app/services/the_ingestor.rb
```diff
@@ -8,22 +8,26 @@ def ingest
...
-    @document.transaction do
-      @chunks.each(&:save!)
-    end
-    @document.chunked!
-
-    @document.embedding!
-    @chunks.each_with_index do |chunk, idx|
-      Rails.logger.info("Embedding chunk #{chunk.id} (#{idx}/#{@chunks.count})...")
+    index = 0
+    chonker.each do |c|
```
nickthecook, line 12:
```suggestion
    chonker.each_with_index do |c, index|
```

app/services/chonker.rb
oxaroky02: Nope. Gone.

app/services/the_ingestor.rb
```diff
@@ -8,22 +8,26 @@ def ingest
...
+
+      chunk = Chunk.new(document: @document, content: c)
+      chunk.save!
+
+      Rails.logger.info("Embedding chunk #{chunk.id} (#{index})")
       embedding = embedder.embed(chunk.content)
       ids = chromadb.add_documents(@collection_id, [chunk.content], [embedding])
       chunk.update!(vector_id: ids.first)
+
+      index += 1
```
nickthecook, line 27: Remove if you use the suggestion about `each_with_index`.

app/services/the_ingestor.rb
oxaroky02: Since `Chonker` is not implementing `Enumerator`, I thought this wouldn't work. I.e. Only `#each` is available from `Chonker`.

app/services/chonker.rb
oxaroky02: Nope. I moved `private` up a little to _also_ make `#chunks` private. This remains private.

nickthecook: Oops!

app/services/the_ingestor.rb
nickthecook: True. You could just implement `each_with_index` to avoid this C code. ;)

oxaroky02: Done ... for now. :-)

PR #32, opened 2024-05-09
Support provisioing of modelConfigs & modelServers:

LeShaunJ:
- Adds `provisioned` column to both tables.
- Adds `enabled` column to both table.
- If provisioned entities are removed from configuration, said entities are disabled, rather than deleted (_and no longer consider "provisioned"_).
- Fixes a potential `compose` bug where `command:` now expects an excutable expression in the for of an array.

### DIFF
--- a/db/migrate/20240509190359_add_provisioning_columns_to_model_tables.rb
+++ b/db/migrate/20240509190359_add_provisioning_columns_to_model_tables.rb
@@ -1,0 +1,9 @@
+class AddProvisioningColumnsToModelTables < ActiveRecord::Migration[7.1]
+  def change
+    add_column :model_configs, :provisioned, :boolean, default: false, if_not_exists: true
+    add_column :model_servers, :provisioned, :boolean, default: false, if_not_exists: true
+
+    add_column :model_configs, :enabled, :boolean, default: true, if_not_exists: true
+    add_column :model_servers, :enabled, :boolean, default: true, if_not_exists: true
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -7,6 +7,7 @@
 #   ["Action", "Comedy", "Drama", "Horror"].each do |genre_name|
 #     MovieGenre.find_or_create_by!(name: genre_name)
 #   end
+require 'json'
 
 # USERS
 #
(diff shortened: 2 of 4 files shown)


db/migrate/20240509190359_add_provisioning_columns_to_model_tables.rb
```diff
@@ -0,0 +1,9 @@
+class AddProvisioningColumnsToModelTables < ActiveRecord::Migration[7.1]
+  def change
+    add_column :model_configs, :provisioned, :boolean, default: false, if_not_exists: true
```
nickthecook, line 3: Why do you need `if_not_exists` here?

db/seeds.rb
```diff
@@ -60,34 +61,84 @@
...
+    model_server: "localhost",
+    model: "gemma:7b",
+    temperature: 0.2,
+  },
+  {
+    name: "all-minilm",
+    model_server: "localhost",
+    model: "all-minilm",
+    embedding: true,
+  },
```
nickthecook, line 91: Let's never use this one. It only supports up to 128 characters when doing embeddings, so any longer chunks are not properly searched.

db/seeds.rb
```diff
@@ -60,34 +61,84 @@
...
+    name: "nomic-embed-text",
+    model_server: "localhost",
+    model: "nomic-embed-text",
+    embedding: true,
+  }
+]
+
+provisioned_model_servers = JSON.parse(ENV.fetch("PROVISIONED_MODEL_SERVERS") {
+  Rails.env == "development" ? JSON.generate(devel_model_servers) : '[]'
+})
```
nickthecook, line 102: To make it more readable and avoid converting a hash to JSON and then parsing it back into a hash, maybe something like this?

```suggestion
provisioned_model_servers = if ENV["PROVISIONED_MODEL_SERVERS"].present?
  JSON.parse(ENV["provisioned_model_servers"])
elsif Rails.env == "development"
  devel_model_servers
else
  []
})
```

db/migrate/20240509190359_add_provisioning_columns_to_model_tables.rb
```diff
@@ -0,0 +1,9 @@
...
+    add_column :model_servers, :provisioned, :boolean, default: false, if_not_exists: true
+
+    add_column :model_configs, :enabled, :boolean, default: true, if_not_exists: true
+    add_column :model_servers, :enabled, :boolean, default: true, if_not_exists: true
```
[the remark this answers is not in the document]
nickthecook: Oh, this was the flag to avoid deleting the model from the DB.

Can you use `available` instead? When we're autodetecting stuff from Ollama I think we'll use the same column, and `available` is more appropriate for that case while still working for this one, IMO.

db/seeds.rb
```diff
@@ -60,34 +61,84 @@
...
+provisioned_model_servers_names = provisioned_model_servers.map { |ms| ms["name"] }
+provisioned_model_config_names = provisioned_model_configs.map { |mc| mc["name"] }
+
+ModelConfig.where(provisioned: true).where.not(name: provisioned_model_config_names).update(enabled: false, provisioned: false)
+ModelServer.where(provisioned: true).where.not(name: provisioned_model_servers_names).update(enabled: false, provisioned: false)
+
+provisioned_model_servers.each do |config|
+  puts "provisioning model server for `#{config["name"]}` ..."
+
+  ModelServer.find_or_create_by!(name: config["name"]).update(
```
nickthecook, line 118: This is tricky, because it only works if all the attributes in the `update` are optional. Looks like they all are, but they shouldn't be!

I think it's okay for now, but I'll need to make all these fields mandatory (e.g. no point in allowing a ModelServer without a `url`) and do something fancier at that point.

It's more common to:

```ruby
ModelServer.find_or_create_by!(name: "...") do |model_server|
  model_server.url = "..."
  ...
end
```

but you want to update all the attributes whether the model was created or not, so that wouldn't work for you.

db/migrate/20240509190359_add_provisioning_columns_to_model_tables.rb
LeShaunJ: You know what, I'd blindly added that as the obvious solution, despite the odd fact that no other migrations needed it. Thinking about it now, it's because I'd initially changed `schema.rb` itself to try things out. In fact, said changed are still in this PR actually; so, I'll remove them.

I'd sort of misinterpreted (_overestimated?_) `db:migrate`. I had this idea that it was implicitly idempotent, when it actually tracks the migrations so it knows never to run them again. While I feel like there's a lane for that to go sideways, and think idempotent migrations are ultimately (_if not, overly_) safer; but, I'll remove the `if_not_exists` as well.

db/seeds.rb
LeShaunJ: Oh right! I forgot to revisit those lol I knew you wouldn't like them

db/migrate/20240509190359_add_provisioning_columns_to_model_tables.rb
oxaronick: `db:migrate` is smart enough to only run migrations that haven't been run on the DB before. You can always assume your migration is being run once.

_Unless_ someone does a `db:rollback`, in which case Rails will try to undo your changes. `add_column` becomes `drop_column`, etc. Then, the next `db:migrate` will run your migration again on that database.

db/seeds.rb
```diff
@@ -60,34 +61,84 @@
...
-  ModelConfig.find_or_create_by!(name: "gemma:7b", model_server: ModelServer.first) do |mc|
-    mc.model = "gemma:7b"
-    mc.temperature = 0.2
-  end
-
-  ModelConfig.find_or_create_by!(name: "all-minilm", model_server: ModelServer.first) do |mc|
-    mc.model = "all-minilm"
-    mc.embedding = true
-  end
+devel_model_servers = [
```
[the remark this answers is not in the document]
LeShaunJ:
```suggestion
dev_model_servers = [
```

db/seeds.rb
LeShaunJ: Yeah, I forgot to ask about the former.

As for the latter, that was my dilemma for like half a day. I'm honestly baffled as to why `do |model_server|` can't be called in either scenario. If you think about it, by the time we reach that point the underlying actions are the same—update an existing record:

1. **Create**: We couldn't find the record; so we've started the process of created one. Set any mandatory fields; and then, we'll call `.save` for you.
2. **Find**: We found the record. We'll update those fields you set during creation and call `.save` for you.

And it could still return the record, as it already does. Anywho, nothing I can do there. That said, I guess the appropriate way to do this would be something like:

```ruby
record = ModelServer.find_or_create_by!(name: "...") do |model_server|
  model_server.url = "..."
  ...
end

record.update(url: "...", ...)
```

Which is.... fine, I guess lol

db/migrate/20240509190359_add_provisioning_columns_to_model_tables.rb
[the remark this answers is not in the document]
LeShaunJ: Oh crap, forgot to change this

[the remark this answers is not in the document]
LeShaunJ: And done.

db/seeds.rb
LeShaunJ: Will leave this for now, as I don't think like the idea of repeating the `create_by` code in the form of an `update`. There's a better, more abstract way to do this, I'm sure.

db/seeds.rb
```diff
@@ -7,6 +7,7 @@
 #   ["Action", "Comedy", "Drama", "Horror"].each do |genre_name|
 #     MovieGenre.find_or_create_by!(name: genre_name)
 #   end
+require 'json'
```
oxaronick, line 10: Did you need this? Rails should auto-load this library.

db/seeds.rb
```diff
@@ -60,34 +61,88 @@
...
-    mc.model = "nomic-embed-text"
-    mc.embedding = true
-  end
+ModelConfig.where(provisioned: true).where.not(name: provisioned_model_config_names).update(available: false, provisioned: false)
+ModelServer.where(provisioned: true).where.not(name: provisioned_model_servers_names).update(available: false, provisioned: false)
+
+provisioned_model_servers.each do |config|
+  puts "provisioning model server for `#{config["name"]}` ..."
 
+  ModelServer.find_or_create_by!(name: config["name"]).update(
```
oxaronick, line 120: This is nice, but has the same problem as the original that if any fields other than `name` are required (and they really should be, I'll fix that) this won't work.

In cases like this, where you want it to find OR basically call `.new` (which won't try to save to the DB and so won't invoke validations), use `find_or_initialize_by` chained with an `update`.

db/seeds.rb
LeShaunJ: This is done

db/seeds.rb
LeShaunJ: I simplified this. It holds no hands and just assumes a given `hash` has the appropriate fields. That, plus the use of `update!` will offer some informative stack traces should the end user miss anything

PR #38, opened 2024-06-05
Use `pdftotext` to parse PDF files

oxaroky02:
The `pdftotext` tool handles PDFs well, and deals with rotated pages well. 

- Run `ops up` to get it into `brew` or install `poppler` yourself
- The `Dockerfile` installed `poppler-utls` to ensure it gets into the built images

### DIFF
--- a/app/services/parsers/pdf.rb
+++ b/app/services/parsers/pdf.rb
@@ -6,34 +6,13 @@ module Parsers
       @document = document
     end
 
-    def pages
-      reader.pages.size
-    end
-
-    def page(index)
-      reader.page(index).text
-    end
-
     def text
-      @text ||= begin
-        text = ""
-        reader.pages.each do |page|
-          text << page.text
-        end
-
-        text
-      end
-    end
-
-    private
-
-    def reader
-      @reader ||= begin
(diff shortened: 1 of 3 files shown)


app/services/parsers/pdf.rb
```diff
@@ -6,34 +6,13 @@ def initialize(document)
...
-        io = StringIO.new
-        io.puts(@document.contents)
-
-        PDF::Reader.new(io)
-      end
+      # NOTE: Using -raw opt causes text to be broken up a lot; but not using raw
+      #       may cause tables to be "pretty" in text which may not be ideal for chunking.
+      #       Not specifying works best for rotated pages, so doing that for now
+      cmd = 'pdftotext - -'
+      txt, = Open3.capture2(cmd, stdin_data: @document.contents, binmode: true)
```
nickthecook, line 14: What do you think about capturing stderr and logging it if present?

oxaroky02: Done.

oxaronick (approved):
Poppler does a much better job parsing the few PDFs I've tried.

PR #40, opened 2024-06-17
Text chunking using recursive text splitting for plain text and markdown

oxaroky02:
- Use the `baran` gem which implements character splitters based on Langchain's text splitters
- Define a recursive text splitting chunker with customizable separator rules
- Modify `Text` parser to use recursive text splitting
- Define `CommonMark` parser derived from `Text` with custom separator rules to account for markdown and embedded HTML tables
- Modify `Docx` parser to subclass `CommonMark` and tweak pandoc invocation to not word-wrap long lines.
- Modify `Parsers#parser_for` to select `CommonMark` parser for `.md` files
- Modify the default chunking size/overlap to bigger defaults of 1000 and 200 characters respectively, which yield better chunks, especially with the new text splitter

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -101,3 +101,5 @@ gem "rubocop-rspec", "~> 2.28"
 gem "pagy", "~> 8.0"
 
 gem "motor-admin", "~> 0.4.26"
+
+gem "baran", "~> 0.1.12"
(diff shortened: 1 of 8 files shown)


oxaronick:
These chunks are readable!

image

oxaronick:
Before I dive in, is it expected that PDF files would still use simple word splitting?

oxaronick (commented):
Looking good, and not much code required to make that work.

Issue #1, opened 2024-03-16
Include previous messages in context


nickthecook:

oxaronick:
Fixed in 1.7.0.

PR #41, opened 2024-06-19
Return parsed chunks as ChunkRecords

oxaroky02:
- Added `Parser::ChunkRecord` which holds two properties: `content` and `key`. 
  - When creating instances, key is optional
  - Records without a key return content as the key
- Modified the two chunkers to return arrays of `ChunkRecord` instead of just the content strings
- Tweaked `TheIngestor` to handle `ChunkRecord` by
  - creating embedding using `#key` from the record
  - storing the `#content` for insertion into the context for user questions

This sets us for exploring future chunking methods where the key (and therefore embedding) has more information than the associated "content". No functional change to current chunkers.

### DIFF
--- a/app/services/the_ingestor.rb
+++ b/app/services/the_ingestor.rb
@@ -8,18 +8,19 @@ class TheIngestor
     ensure_collection_exists
 
     @document.chunking!
-    chonker.each_with_index do |c, index|
+    chonker.each_with_index do |crec, index|
       if index.zero?
         # Temporary hack since parser is still a brute force chunker right now.
         @document.chunked!
         @document.embedding!
       end
 
-      chunk = Chunk.new(document: @document, content: c)
+      chunk = Chunk.new(document: @document, content: crec.content)
       chunk.save!
 
       Rails.logger.info("Embedding chunk #{chunk.id} (#{index})")
-      embedding = embedder.embed(chunk.content)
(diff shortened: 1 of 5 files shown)


oxaronick:
This makes the `key` transient. Should it be generated on the fly and then discarded, or should Archyve retain it as a new field in `Chunk`?

Retaining it would help with debugging, especially if we're trying out different algorithms for generating the key based on the document or content.

oxaroky02:
> This makes the `key` transient. Should it be generated on the fly and then discarded, or should Archyve retain it as a new field in `Chunk`?

Yes, ideally, but two reasons:

1. For irrational (fear?) reasons I wanted to make this change _before_ making the model change/migration. 😬 
2. Also, before that I wanted "try out" and see what what other `ChunkRecord` changes might be needed. 

It's not as end-to-end, I agree. And I think adding the "key" field (better name needed, maybe) to `Chunk` model is OK to do as part of this. I can get over #1. 😄

app/services/the_ingestor.rb
```diff
@@ -8,18 +8,19 @@ def ingest
     ensure_collection_exists
 
     @document.chunking!
-    chonker.each_with_index do |c, index|
+    chonker.each_with_index do |crec, index|
```
oxaronick, line 11: What about `content_record`, or even `record`?

PR #42, opened 2024-06-24
Adding column to keep track of the embedding content per Chunk

oxaroky02:
- `Chunk` model has `embedding_content` column/field
- View for chunks updated top show `#embedding_content`


app/services/the_ingestor.rb
```diff
@@ -9,33 +9,43 @@ def ingest
...
+  end
+
+  def document_embedding!
+    # Temporary hack since parser is still a brute force chunker right now.
+    @document.chunked!
+    @document.embedding!
+  end
+
+  def prepare_and_embed_chunk(chunk_record, index)
+    document_embedding! if index.zero?
```
oxaronick, line 37: Could you do this once, before the outer `each_with_index`?

app/services/the_ingestor.rb
```diff
@@ -9,33 +9,43 @@ def ingest
...
+    chunk = Chunk.new(
+      document: @document, content: chunk_record.content,
+      embedding_content: chunk_record.embedding_content)
+    chunk.save!
+
+    # Rails.logger.info("Embedding chunk #{chunk.id} (#{index})")
```
oxaronick, line 43: Did you mean to commit this commented out?

app/services/the_ingestor.rb
oxaroky02: I call `#chunking!` outside the loop and then this inside to track `#chunked!` because it's possible that `chonker.each_with_index` could fail after "chunking" but before "chunked."

Pedantic? Yes.

oxaroky02: Actually, I have refactor for this, coming shortly.

nickthecook (approved):

PR #44, opened 2024-07-02
[PATCH] ensure provisioning fails safely

LeShaunJ:
Heavily refactors original implementation:
* Introduces `provision_from_env(schema, env_var, identifier='name', &block)`:
  * `schema` is a `class ApplicationRecord`.
  * `env_var` is the name of the environment variable the end-user will use to provision records.
  * `identifier` is the optional column name that will be used to match existing records.
  * `&block` can return an optional default provisioning value for `dev` environments.
* Will succeed; or otherwise, _skip_ execution if:
  * `ApplicationRecord` schema is missing the columns, `available` & `provisioned`.
  * JSON fails to parse or resulting value is not an `Array` type.
* Provisioning of specific records:
  * Is _skipped_ if the `identifier` key is missing from the JSON object.
  * Filters out keys that do not correspond to the column names of the `ApplicationRecord`.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -49,66 +49,72 @@ end
 
 # PROVISIONING
 #
-dev_model_servers = [
-  {
-    "name" => "localhost",
-    "url" => "http://localhost:11434",
-    "provider" => "ollama",
-  }
-]
+def provision_from_env(schema, env_var, identifier='name', &block)
+  known_columns = schema.column_names
+  unless ['available', 'provisioned'].all? { |required| known_columns.include? required }
+    return puts("WARNING: #{schema.table_name} does not support provisioning. Skipping...")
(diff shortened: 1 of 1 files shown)


oxaronick:
This looks a bit like you're trying to protect the seeds from changes in the app. Since the seeds are in the same codebase, usually we'd skip defensive coding and just update the seeds when we update the app.

I do forget to update seeds sometimes, and if you are working mostly with seeds and not with the rest of the app maybe this makes sense. However, I worry that, long-term, this kind of complexity will make the seeds hard to understand. Even prior to this MR, the seeds are already more complex than most Rails apps' seeds.

How about we pair on this and see if we can accomplish what you are trying to do in a more customary way for a Rails app?

PR #47, opened 2024-07-24
Add Azure OpenAI client

oxaroky02:
> Updated to reflect various changes based on feedback; added some tests.

## TLDR;

- This is big PR since adding Azure Open AI support also meant making all the rest of Archyve support _any LLM client_, which in turn meant removing assumptions about Ollama client-specifics.
- The Azure OpenAI client support _chat,_ _embedding_ and _completion_ requests.

The detailed description is organized as follows.

- Model changes
  - `ModelServer`
  - `ModelConfig`
  - `ApiCall`
- LLM Clients
  - Common
  - OpenAI Client
  - Ollama Client
  - Base Client
- Services
  - Helpers
  - Common Client Support
  - API

### ModelServer

- Tweaked the `provider` enumeration, renaming `openai` to `openai_azure` and setting it up with a naming prefix so that we can query a model server with `server.provider_ollama?` and `server.provider_openai_azure?`.
- Added (via migration) optional `api_key` field for use with external servers that require API tokens.
- Added validation support methods `api_key_required?` and `api_version_required?` implemented based on a server's provider, both of which return true if the provider is `openai_azure`.
- Added validation to enforce presence of `api_key`.

### ModelConfig

- Added two fields (via migration), an optional `api_version` and an optional `model_server` reference.
- Added validation support method `api_version_required?` which calls the same method for a model server if present
- Added validation to enforce presence of `api_version`

### ApiCall

- Added `from_faraday` support method to create an API call record based on Faraday-middleware based request/responses.
- Used by the Open AI client via a Faraday request interceptor.

> Caveat: When using Open AI "streaming", we can't retrieve the response body via the interceptor.

### Common

- Tweaked to `#chat` (in all clients) to receive `Message` and handle the chat history conversion within the client

### OpenAI Client

- Added the `ruby-openai` gem to the project's bundler
- Re-implemented the `openai/` client API as follows under the `Openai::` module as follows:
  - `Client` base class for non-Azure-specific use of the `ruby-openai` gem, including integration with Faraday's `interceptor` middleware flow to track request/response details
    - **Update**: removed the mixins and put the `chat_request` and `complete_request` and `embedding_request` methods into the `Client` where they belong for clarity
  - `ChatMessageHelper` to help process `Message` for chat history formatting for server-specific requests
  - `AzureClient` implementation sub-class for all the Azure-specific aspects of the client. 
    - **Update**: Renamed some internal method names to use "connection" when referring to the underling `ruby-openai` "client" to avoid ambiguity with the use of "client" in Archyve.

### Ollama Client

- Modified `#chat` to use the OpenAI `ChatMessageHelper` for the chat history, since Ollama API is compatible with OpenAI.

### Base Client

- Added optional `api_version` property
- Tweaked `client_class_for` factory method to return appropriate client class based on the provider

### Helpers

- Started a `Helpers` module
- Added a `ModelClientHelper` helper that is shared by various services, consolidating common functionality when instantiating an LLM client based on a `ModelConfig`.
  - **Update**: Converted from mixin to proper helper instantiated by services.
- Removed any Ollama client-specific references

### Common Client Support

- Refactored the following service classes to leverage the shared `ModelClientHelper` (**update**: replaced mixin with helper instances) for the common functionality to be able to retrieve an LLM client based on the model configuration:
  - `ResponseStreamer`, 
  - `Embedder` and 
  - `SummarizeMessage`
- Removed any Ollama client-specific references, including letting the clients deal with the `Message` processing for chat history.
- Tweaked the `TheIngestor`, `ReplyToMessage` and `Search` services to reflect the changes in the refactored service classes

### API

- Tweaked the `ModelLoader` to support the model server's API key

## Testing

- Refactored the `ollama/chat_spec` to instead test the Open AI `ChatMessageHelper` shared by both clients
- Added an Ollama and Azure Open client test specs to test the three method `#embed`, `#chat`, and `#complete`
- Put in a `spec_helper.rb` config check to exclude tests tagged with `:az_openai` unless both `AZURE_OPENAI_API_KEY` and `AZURE_OPENAI_URI` env variables have been set.


app/lib/llm_clients/ollama/client.rb
```diff
@@ -9,8 +9,9 @@ def complete(prompt, &)
         stream(request, &)
       end
 
-      def chat(chat, &)
-        request = helper.chat_request(chat.chat_history)
+      def chat(message, &)
+        # Ollama chat protocol matches the one for OpenAI, so use the same helper
```
oxaronick, line 13: Ollama's implementation must have been "heavily inspired by" OpenAI's.

app/lib/llm_clients/openai/azure_client.rb
```diff
@@ -0,0 +1,32 @@
...
+        az_client(@model)
+      end
+
+      def embed_client
+        az_client(@embedding_model)
+      end
+
+      def az_client(model_name)
+        @az_client ||= OpenAI::Client.new(
+          access_token: @api_key, #ENV.fetch("AZURE_OPENAI_API_KEY"),
```
oxaronick, line 22: Remove comment?

I agree we shouldn't use an env var, since a single API key doesn't really fit with the ModelServer concept.

app/lib/llm_clients/openai/chat_request_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+                      end
+                      num_tokens += 1
+                      unless (str = chunk.dig("choices", 0, "delta", "content")).nil?
+                        current_batch << str
+                        current_batch_size += 1
+                        if str.ends_with?("\n") || str.ends_with?("}") || current_batch_size >= @batch_size
+                          Rails.logger.debug { "==> #{current_batch}" }
+                          yield current_batch
+
+                          current_batch_size = 0
```
oxaronick, line 31: Ruby's `String#length` method is BigO(1) (unless the string is multi-byte encoded), so you could dispense with this counter and just use `current_batch.length` if you want.

app/lib/llm_clients/openai/client.rb
```diff
@@ -1,77 +1,74 @@
...
 module LlmClients
   module Openai
-    class Client < Client
-      def call(prompt)
-        @stats = new_stats
-        request = request(prompt)
+    class Client < LlmClients::Client
+      include ChatRequestHelper
+      include EmbeddingRequestHelper
+      include CompletionRequestHelper
```
oxaronick, line 8: Do these need to be `include`d, or could you use them as collaborators instead? E.g.:

```ruby
def chat_request_helper
  @chat_request_helper ||= ChatRequestHelper.new(...)
end
```

app/lib/llm_clients/openai/client.rb
```diff
@@ -1,77 +1,74 @@
...
-              @stats[:tokens] += 1
-              content += message if message
-            end
+      def complete(prompt, &)
+        complete_request(prompt, &)
+      end
 
-            yield content
+      def chat(message, &)
+        chat_request(ChatMessageHelper.new(message).chat_history, &)
```
oxaronick, line 17: Oh, you're calling `new` anyway; do you need the `include`s?

app/lib/llm_clients/openai/client.rb
```diff
@@ -1,77 +1,74 @@
...
-            calculate_stats
-          end
-        end
+      def embed(content)
+        embedding_request(content)
+      end
+
+      # Callback for instrumenting request via Faraday middleware used by OpenAI API gem
+      def instrument(_name, env)
+        tmp = yield
```
oxaronick, line 26: Would calling this `response` be more clear?

app/lib/llm_clients/openai/client.rb
```diff
@@ -1,77 +1,74 @@
...
 
-      def extract_message(response_string)
-        response_hash = JSON.parse(response_string)
-        message = response_hash["choices"].first["delta"]["content"]
-        Rails.logger.info("<== '#{message}'")
-
-        message
+      # Create an ApiCall based on a Faraday environment for this client
+      def api_call_for(env)
+        ApiCall.from_faraday(
```
oxaronick, line 44: Nice!

app/lib/llm_clients/openai/client.rb
```diff
@@ -1,77 +1,74 @@
...
+            http_method: env[:method].downcase,
+            url: env[:url],
+            headers: clean_headers(env[:request_headers]),
+            body: env[:request_body],
+          },
+          response: {
+            headers: env[:response_headers],
+            status: env[:response].status,
+            body: env[:response].body,
+          },
```
oxaronick, line 56: It looks like you're doing some of the unpacking from a Faraday env here, and doing the rest in `ApiCall`. Since the method is called `from_faraday`, would it be simpler if you just passed `env` instead of `request` and `response` and left all the work to `ApiCall.from_faraday`?

app/lib/llm_clients/openai/azure_client.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Openai
+    class AzureClient < Client
+      private
+
+      def client_provider
+        "openai_azure"
+      end
+
+      def chat_client
+        az_client(@model)
```
oxaronick, line 13: This isn't really a "client", is it? It's more of a ClientFactory. Right now, one would call `AzureClient.new(...).chat_client` to get a client, right?

What do you think about leaving this work to class methods, and just making the Client class be a client?

E.g.:

```ruby
class << self
  def chat_client
    new(endpoint, ...)
  end

  def embedding_client
    new(...)
  end
end
```

oxaronick: Another option would be to just implement another class for this factory-type work, I'm not stuck on class methods. I just think logic to build clients is better outside the client instance itself.

app/services/embedder.rb
```diff
@@ -1,27 +1,9 @@
 class Embedder
-  def initialize(embedding_model, traceable: nil)
-    @embedding_model = embedding_model
-    @traceable = traceable
-  end
+  include Helpers::ModelClient
```
oxaronick, line 2: I'd avoid mixins here as well, if possible.

In general, they make testing harder and make your class behave in ways that are not obvious from the code in the class itself. It's better to have collaborators than mixins.

app/services/search/search_n.rb
```diff
@@ -2,7 +2,7 @@ module Search
   class SearchN < Base
     def initialize(collections, num_results: 10, traceable: nil, include_irrelevant: false)
       @collections = collections
-      @num_results = num_results
+      @num_results = num_results || 10
```
oxaronick, line 5: 10 is already the default; do you need this?

oxaronick:
I'm going to stop there for now. In general, I think it's great! Just some concerns about how the code is structured.

There are no tests, but I can contribute those afterwards. Archyve isn't near 100% test coverage yet, but there are 100-odd tests in there now, and I'm trying to test new code.

app/lib/llm_clients/openai/chat_request_helper.rb
oxaroky02: I based this on how the batch size was implemented in the `ollama` client, where it looks like the batch size is "token count" vs number of characters. See lines 55-59 in `llm_clients/ollama/client.rb`.

So decision here could go two ways:
1. It should have been character count, in which case I'll switch to use `#size`
2. It should be token count, in which case I can refactor to rename things to better reflect intent.

Thoughts?

app/lib/llm_clients/openai/chat_request_helper.rb
```diff
@@ -0,0 +1,50 @@
...
+                          current_batch = ""
+                        end
+                      end
+                    end,
```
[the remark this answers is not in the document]
oxaroky02: I'm using `ruby-lsp` and auto formatting. I checked to confirm it's working and I keep getting back this layout. Should it look different? I tried manually changing and when I save it reverts to this.

app/lib/llm_clients/openai/client.rb
oxaroky02: Hmm, I think this is due my misunderstanding on my side, I was putting "supporting methods" in separate files but calling them "helper" which is incorrect. For the time being I'll pull the methods into the `client.rb` where they typically would exist.

app/lib/llm_clients/openai/client.rb
oxaroky02: See above re helpers. I was using two patterns but calling them both helpers. In this case this is a helper proper, and I think this usage is correct _and_ unrelated to the various modules I was including.

app/lib/llm_clients/openai/client.rb
oxaroky02: Yes it would. :-)

app/lib/llm_clients/openai/client.rb
oxaroky02: Originally I did have an implementation where I passed the `env` into `from_faraday`. However I also wanted to make some changes to the contents in `env` that would be "client" specific. E.g. cleaning the headers to mask the API key, where different client protocols might use different naming. Also, I didn't want to edit `env` in place. Eventually I just landed on this as cleanest approach. 

So, why call it `from_faraday`? The values are still specific to Faraday, so if there was any processing in `ApiCall` it helps to know that in case there are type specifics for the values.

app/lib/llm_clients/openai/azure_client.rb
oxaroky02: Ah, just bad naming. I was using `client` in the `ruby-openai` sense but since we're defining "LLM client" I should really using "connection". These represent the under-the-hood connectivity.

app/services/search/search_n.rb
oxaroky02: I ran into an issue where there was a Setting was being passed in, but it was blank. So the default in the method declaration didn't kick in. It was due to bad data in the DB, so maybe should let it fail?

app/services/embedder.rb
oxaroky02: Makes sense. Done.

oxaroky02:
> > There are no tests, but I can contribute those afterwards. Archyve isn't near 100% test coverage yet, but there are 100-odd tests in there now, and I'm trying to test new code.
> 
> Yep. I did want to include tests but I ran into some trouble running the tests. I promise to start adding testing in separate PRs.

OK, figured out my problem with running tests; can run tests now, and found that the old `ollama` chat test failed due to my refactoring. So I've fixed that so at least all the current tests pass again. Sorry about that.

app/lib/llm_clients/openai/chat_request_helper.rb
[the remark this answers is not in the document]
oxaronick: Hmm... there are rules in .rubocop.yml that should correct that. I'm getting autoformatting too, and it's not doing that with my code.

Maybe your plugin is not running rubocop with `bundle exec`?

app/lib/llm_clients/openai/chat_request_helper.rb
oxaronick: It's fine.

app/lib/llm_clients/openai/client.rb
oxaronick: I was thinking not so much of separate files - separate files makes sense, but rather than `include` you could just instantiate the helper. It's "is a" vs. "has a", and "is a" can make things more complicated.

I'll take a look at what you have.

app/lib/llm_clients/openai/azure_client.rb
oxaronick: It looks like now the "Client" is the factory, and the "connection" is the client. I was just thinking this class called Client wasn't actually a client in any meaningful way, so it should be called something else.

I think this is an issue in general. Clients are hard to create because of all the parameters you either a) need to have, or b) need to derive. I added a ModelLoader class for the API controllers because it was a pain.

Maybe we should embrace the factory pattern and just call these classes `ClientFactory` or `ClientHelper` or something.

Ah, I think what throws me off about this is that AzureClient _is a_ Client, _and_ it _has a_ Client. Does that seem odd to you?

app/services/search/search_n.rb
oxaronick: Ah, right.

If you want it to be valid to explicitly pass `nil` to this method and still get the default value, then I'd drop the param default and use the default inside the constructor.

I guess if you _also_ want to not pass the arg at all, you could default `num_results` to `nil` - then you don't need to provide, and if you don't the default is used. But you can provide it as `nil`, and the default is also used.

app/lib/llm_clients/openai/chat_request_helper.rb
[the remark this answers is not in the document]
oxaroky02: Adding `"rubyLsp.linters": ["bundle exec rubocop"],` to the settings fixed the formatting. Done.

app/lib/llm_clients/openai/azure_client.rb
oxaroky02: From Archyve's point of view `AzureClient` is the LLM client used by the app. Under the hood we're using a gem who also happens to call their "entity" a "client". To deal with the confusion I'm calling it a "connection" now. So the LLM client uses an underlying connection. :-) I think it's OK.

app/lib/llm_clients/openai/client.rb
oxaroky02: I was using mixins but calling it a helper. I don't want a helper since I have to do a whole bunch of parameter passing which in this case makes no sense. These aren't helpers, now that I properly understand the pattern. So I've moved the three methods into `client.rb`, which matches what you have in the `ollama/client.rb` for now.

app/lib/llm_clients/openai/azure_client.rb
oxaronick: nm, AzureClient is instantiating `OpenAI::Client`, not `Openai::Client`. My mistake.

PR #52, opened 2024-07-31
Refactor to separate chunkers from parsers so any parser can use any chunker

oxaroky02:
> Note: It looks like a lot but the actual change isn't that complicated; however many files affected as I moved things around.

### Background

Prior to this PR, we had document parsers that were based on one specific chunker; i.e. the PDF parser was based on the basic chunker while the text/md/docx parsers were based on recursive splitter, and neither could really use a different chunker. 

### Changes

- `Chonker` is gone. 😄 
- Defined `Chunkers` module with ... 
  - Tweaked `CHUNKING_METHODS` as an array of methods each with an `id` and `name`.
  - Added `InputType` enumeration (not Rails) 
  - Added `#chunker_for(chunking_profile)` to find and create a chunker based on the profile (and it's method)
- Separated chunkers under the `Chunkers` module in the `services/chunkers/` folder. 
- Two chunkers implemented as collaborators rather than mixins: 
  - `BasicCharacterChunker` which is the original regex-based splitter
  - `RecursiveTextChunker` which is the recently added chunker that uses separator sets to assist with prioritized recursive splitting using the `baran` gem.
- Each chunker supports `#text(text, text_type)` method so the parser can specify the input type when chunking
- The parsers have been refactored to not mixin the chunking method, instead the base `Text` parser supports instantiating the appropriate chunker based on the document's chunking profile.
- Add tests

### Outcome

When ingesting a document into a collection, the user can select either chunking method for any of the supported document types.


oxaroky02:
> Converting to draft ... I don't like how a chunker is being instantiated. Hold on.

Done. Simpler now too.

app/services/chunkers.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  include Enumerable
+
+  module InputType
+    PLAIN_TEXT = 0
+    COMMON_MARK = 1
+  end
+
+  CHUNKING_METHODS = [
+    { id: :recursive_split, name: "Recursive Split" },
```
nickthecook, line 12: What do you think of just having the snake case name and using '.titleize' (and eventually translations) to get the view string?

app/services/chunkers.rb
```diff
@@ -0,0 +1,35 @@
...
+    { id: :basic, name: "Basic" },
+  ].freeze
+
+  def self.chunker_for(chunking_profile)
+    chunking_method = chunking_profile.method.to_sym
+    if chunking_method == :bytes
+      # Mapping :bytes method for backwards compatibility since
+      # the DB will have existing chunking profiles that we might want to "reprocess"
+      Rails.logger.info("Backwards compatibility - mapping :bytes to :recursive_split chuking method")
+      chunking_method = :recursive_split
```
nickthecook, line 22: Is :bytes not more like :basic?

app/services/parsers/text.rb
```diff
@@ -1,14 +1,35 @@
...
       @text ||= @document.contents
     end
+
+    private
+
+    def text_type
+      Chunkers::InputType::PLAIN_TEXT
+    end
+
+    def chunker
```
nickthecook, line 31:
```suggestion
    def chonker
```

nickthecook (commented):
Minor questions; looks good!

app/services/chunkers.rb
oxaroky02: Yes and no only because previously we used the `bytes` method and let the parser pick _actual_ the chunking method. This means that prior to this PR, PDF chunking used the basic chunker but the other formats used the recursive split chunker. However with this separation of chunker from parser, I had to pick one or other for backwards compatibility, and I went with the "better" of the two strategies. 

What we could do, ideally, is prompt the user on a "re-process" request if they want to keep the previous profile or change it. 

With that said, I am OK with `bytes` mapping to either; if we have a lot more PDFs in deployed systems, then using `basic` is gonna provide most compatibility. Let me know. Easy fix.

app/services/chunkers.rb
oxaroky02: I considered that; however I wasn't sure if the user-friendly name was always going to map that well to the symbol. While internally we may use a "Recursive Character Splitter", for a user those words mean nothing and we may instead want to say something else. 

That aside, I chose the "array of records" because I wanted to add some more properties. E.g. defaults for chunk size and overlap. In addition, some other chunking methods don't require chunk size or overlap as parameters, or add others.

oxaroky02:
Update: added chunker tests which revealed a bug in how `#chunk` is called in the recursive splitter chunker. +1 for writing tests. 😄

oxaroky02:
Update: added parser tests and data files. (OK, I'm done now. 😄 )

app/services/chunkers.rb
nickthecook: If we're not going to use the user-friendly name in the UI, what is it for?

app/services/parsers/text.rb
```diff
@@ -1,14 +1,30 @@
...
       @text ||= @document.contents
     end
+
+    private
+
+    def text_type
+      Chunkers::InputType::PLAIN_TEXT
+    end
+
+    def chonker
```
nickthecook, line 26: I'm sorry - I was kidding with the suggestion of `chonker`. Feel free to change it back. :D

nickthecook:
LGTM,  but I'll wait a bit in case you want to change `chonker` back.

app/services/parsers/text.rb
oxaroky02: No worries; it's a single private method. No harm.

PR #82, opened 2024-09-18
Add require open3 gem to avoid error

### LINKED ISSUE #81: Missing open3 gem
Adding a pdf document gives me the error below, which I resolved by adding `gem "open3"` to Gemfile. Ruby 3.1.4.
Maybe adding `require "open3"` would also fix it.
```
jobs    | 2024-09-18T21:28:30.248Z pid=59558 tid=ohq WARN: NameError: uninitialized constant Parsers::Pdf::Open3
jobs    |
jobs    |       txt, stderr, status = Open3.capture3(cmd, stdin_data: @document.contents, binmode: true)

mattlindsey:
I reproduced the error in a few different versions of ruby 3, and this fixed it.
Fixes #81

### DIFF
--- a/app/services/parsers/docx.rb
+++ b/app/services/parsers/docx.rb
@@ -1,3 +1,5 @@
+require 'open3'
+
 module Parsers
   # DOCX parser converts word documents into commonmark (with tables) markdown
   class Docx < CommonMark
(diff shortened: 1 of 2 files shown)


nickthecook:
`open3` actually isn't in the `Gemfile.lock` either, so we must just both have it installed globally for something else.

How about adding it to the Gemfile instead, with `bundle add open3`?

nickthecook:
Thanks!

PR #83, opened 2024-09-19
Add default conversation source

mattlindsey:
Avoid form error caused if someone creates a new conversation without any content.  Otherwise, subsequently clicking on the empty coversation causes the form to blow up on NilClass error.


nickthecook:
Makes sense.

Issue #68, opened 2024-09-06
Web Scraping


nickthecook:
Instead of uploading a document, the user should be able to enter a URL.

Archyve needs to:

- find and scrape the main document content
- ideally show the user a preview of the text scraped from the page before actually ingesting the content
- once the user confirms, chunk, embed, and graph the content just like it was a document

Along the way, this will require:

- updating Document so that it works for a web source instead of just an uploaded attachment (preferred to creating a new model and having `Collection#documents` be polymorphic)
- updating the Collection view so it shows web sources along with Documents

mattlindsey:
Since the web page has now been scraped, an enhancement now could be for the chat augmentation to include the URL(and maybe date scraped) that a chunk was derived from in case the llm wants to use it to retrieve different information.  Does that make sense?

PR #86, opened 2024-09-27
Add html parsing and web scraping

mattlindsey:
For issue #68.

Still work in progress, but hoping for feedback.  Try importing a page from a site.  Need to fix tests.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -65,7 +65,7 @@ class DocumentsController < ApplicationController
         f.puts HTTParty.get(params[:link])
         f.rewind
         document.file = f
-        document.filename = file.path #File.basename(f.path)
+        document.filename = File.basename(f.path)
       rescue StandardError
         document.update(state: :errored)
       end
@@ -84,7 +84,7 @@ class DocumentsController < ApplicationController
   end
 
   def document_params
-    params.permit(:file, :link)
+    params.permit(:file, :link, :filename)
   end
 
   def set_document
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -19,7 +19,7 @@ class DocumentsController < ApplicationController
 
     @document = new_document
 
-    ChunkDocumentJob.perform_async(@document.id)
+    ChunkDocumentJob.perform_async(@document.id) unless @document.errored?
 
     respond_to do |format|
       format.turbo_stream do
@@ -59,7 +59,19 @@ class DocumentsController < ApplicationController
 
   def new_document
     document = Document.new(document_params.merge(chunking_profile: @chunking_profile))
-    document.filename = params[:file].original_filename
+    if params[:link].present?
+      begin
+        f = Tempfile.create(['web-', '.html'])
+        f.puts HTTParty.get(params[:link])
+        f.rewind
+        document.file = f
+        document.filename = file.path #File.basename(f.path)
+      rescue StandardError
+        document.update(state: :errored)
+      end
+    elsif params[:file]
+      document.filename = params[:file].original_filename
+    end
     document.collection = @collection
     document.user = current_user
     document.save!
@@ -72,7 +84,7 @@ class DocumentsController < ApplicationController
   end
 
   def document_params
-    params.permit(:file)
+    params.permit(:file, :link)
   end
 
   def set_document
--- a/app/views/documents/_form.html.erb
+++ b/app/views/documents/_form.html.erb
@@ -6,7 +6,10 @@
         <div class="flex-col">
 
           <%= f.file_field :file, class: "w-full p-2 rounded-lg my-2 px-5 py-2 border border-secondary-400 dark:border-secondary-600" %>
-          <%= f.url_field "link", class: "w-full p-2 rounded-lg my-2 px-5 py-2 border border-secondary-400 dark:border-secondary-600" %>
+          <%= f.url_field "link",
+            placeholder: "Enter a URL",
+            class: "w-full p-2 rounded-lg my-2 px-5 py-2 border border-secondary-400 dark:bg-secondary-925 dark:border-secondary-600"
+          %>
           <%= fields_for :chunking_profile do |cp| %>
             <div class="flex w-full justify-between overflow-x-scroll no-scrollbar">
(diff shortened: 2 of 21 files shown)


nickthecook:
Thanks for the PR for such a cool feature. I'll take a look this morning!

app/controllers/documents_controller.rb
```diff
@@ -59,7 +59,19 @@ def start; end
...
   def new_document
     document = Document.new(document_params.merge(chunking_profile: @chunking_profile))
-    document.filename = params[:file].original_filename
+    if params[:link].present?
+      begin
+        f = Tempfile.create(['web-', '.html'])
+        f.puts HTTParty.get(params[:link])
+        f.rewind
+        document.file = f
+        document.filename = file.path #File.basename(f.path)
```
nickthecook, line 68: I had to change `file.path` to `f.path` here to get it to run.

nickthecook (commented):
Looking good! This fits really well into the existing parser structure.

I tried it with the Wikipedia article on LLMs, and it worked.

Seems like it will fit into Document well enough too; are you thinking of making `filename` optional, or maybe renaming it to `title` and trying to populate it with the title of the page?

nickthecook:
Yep, that's my fault. I introduced a bug in Setting that keeps creating duplicate settings. I just ran into it myself.

Go to the admin ui, then Settings, then delete any settings with nil values.

I'll work on a fix.

nickthecook:
I think the root of the problem is that I added validation to Setting so it wouldn't create duplicates, but some duplicates already existed in our databases because of a previous bug.

`rails db:seed` will deduplicate them for you now.

mattlindsey:
Great. Works now.
You can merge this PR if you want. Or if you think there's more to be done, I can keep working on it.

mattlindsey:
Actually don't merge this yet.  
The chunking for plain text and markdown seems to be broken. Looks like chunk_size isn't getting set and may be defaulting to 1000, but I don't know why.  Take a look if you can. Thanks.

mattlindsey:
Fixed this.  Everthing seems good now.

nickthecook:
> Fixed this. Everthing seems good now.

Great! I'll have another look, and comment on code organization/UI stuff this time, since it's out of draft.

app/controllers/documents_controller.rb
```diff
@@ -59,7 +59,19 @@ def start; end
...
-    document.filename = params[:file].original_filename
+    if params[:link].present?
+      begin
+        f = Tempfile.create(['web-', '.html'])
+        f.puts HTTParty.get(params[:link])
+        f.rewind
+        document.file = f
+        document.filename = File.basename(f.path)
+      rescue StandardError
+        document.update(state: :errored)
```
nickthecook, line 70: This should include the exception message in the `error_message` field. The error message will show when the user hovers over the "Error" pill on the document in the UI.

app/controllers/documents_controller.rb
```diff
@@ -59,7 +59,19 @@ def start; end
 
   def new_document
     document = Document.new(document_params.merge(chunking_profile: @chunking_profile))
-    document.filename = params[:file].original_filename
+    if params[:link].present?
+      begin
```
nickthecook, line 63: This might be better in a separate method.

app/controllers/documents_controller.rb
```diff
@@ -59,7 +59,19 @@ def start; end
...
+        f = Tempfile.create(['web-', '.html'])
+        f.puts HTTParty.get(params[:link])
+        f.rewind
+        document.file = f
+        document.filename = File.basename(f.path)
```
nickthecook, line 68: The tempfile names aren't very descriptive. What do you think about grabbing the page title and using that as the document filename?

Only the `file` attribute is managed by ActiveStorage; I don't think `file` and `filename` have to match.

app/views/documents/_form.html.erb
```diff
@@ -1,12 +1,12 @@
...
+<%= form_with url: collection_documents_path(collection), html: { id: "document_form", class: "mt-3 w-full"} do |f| %>
   <div class="flex justify-center">
     <div class="w-full shadow-lg bg-secondary-200 dark:bg-secondary-900 dark:border-secondary-400 border px-6 py-5 mt-3 rounded-lg">
-        <div class="mb-2 text-sm">Add document</div>
+        <div class="mb-2 text-sm">Add document or URL</div>
         <div class="flex-col">
 
           <%= f.file_field :file, class: "w-full p-2 rounded-lg my-2 px-5 py-2 border border-secondary-400 dark:border-secondary-600" %>
-
+          <%= f.url_field "link", class: "w-full p-2 rounded-lg my-2 px-5 py-2 border border-secondary-400 dark:border-secondary-600" %>
```
nickthecook, line 9: This looks good in light mode, but in dark mode the text is light but the bg is white.

Can you theme it like the search bar on the same page?

nickthecook: Also, it would be good if there was some hint for the user as to what the link field was. It does say "Add document or URL" above the file upload box, but maybe if you added placeholder text it would help?

nickthecook (commented):
Looking good! Just a few comments.

app/controllers/documents_controller.rb
mattlindsey: I tried that but it was kinda clunky and a html file on disk can have both a filename and title, so I'd rather add `title` to the documents table and display that instead of filename if it is not blank, if that's ok.

app/views/documents/_form.html.erb
nickthecook: This looks good in dark mode:

```suggestion
          <%= f.url_field "link",
            placeholder: "Enter a URL",
            class: "w-full p-2 rounded-lg my-2 px-5 py-2 border border-secondary-400 dark:bg-secondary-925 dark:border-secondary-600"
          %>
```

nickthecook (approved):
Thanks, @mattlindsey !

nickthecook:
I think somehow the change you made to set the document title was lost.

Any chance you can find that in a branch and push it again?

nickthecook:
Changes to set `Document#error_message` and move the link stuff to a separate method in `DocumentsController` are also lost.

PR #90, opened 2024-10-02
Add context window length property to model config

oxaroky02:
A small step towards some future work where knowing the context window size will be needed to be able to use AI to help with some chunk processing.

### DIFF
--- a/app/models/model_server.rb
+++ b/app/models/model_server.rb
@@ -30,6 +30,19 @@ class ModelServer < ApplicationRecord
     end
   end
 
+  def default_context_window_size
+    if provider_openai_azure?
+      # May be more per model, but we can assume a bigger default
+      8 * 1024
+    elsif provider_openai?
+      # May be more per model, but we can assume a bigger default
+      16 * 1024
+    else
+      # The default for ollama unless overridden by custom Modelfile
+      2 * 1024
+    end
+  end
+
   def make_active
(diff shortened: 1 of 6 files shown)


db/migrate/20241002185332_add_context_window_size_to_model_config.rb
```diff
@@ -0,0 +1,5 @@
+class AddContextWindowSizeToModelConfig < ActiveRecord::Migration[7.1]
+  def change
+    add_column :model_configs, :context_window_size, :integer
```
oxaronick, line 3: What do you think about adding `_bytes` to this so the unit is clear?

app/models/model_server.rb
```diff
@@ -30,6 +30,19 @@ def active_server
     end
   end
 
+  def default_context_window_size
```
oxaronick, line 33: Does this belong in model_server, or model_config?

oxaronick: Never mind... I remember why you put it here.

db/migrate/20241002185332_add_context_window_size_to_model_config.rb
oxaronick: Hm, there is one other column, on ChunkingProfile, that is `size` without `_bytes`. Seems okay.

PR #92, opened 2024-10-04
Fix document new form

mattlindsey:
Fix submit button on document new form.

### DIFF
--- a/app/views/documents/new.html.erb
+++ b/app/views/documents/new.html.erb
@@ -50,7 +50,7 @@
               <div class="flex-1"></div>
               <%= cp.text_field :overlap, value: default_chunk_overlap(default_chunking_method), size: 10, maxlength: 10, class: "rounded-md dark:bg-secondary-925 text-right" %>
             </div>
-            <% end%>
+            <% end %>
           </div>
           <div class="flex py-4 items-center">
             <%= link_to "Cancel", collection_path(@collection), class: "rounded-lg px-4 py-2 text-center text-secondary-950 align-baseline dark:text-secondary-100 block font-medium border border-secondary-700 dark:border-secondary-400 bg-secondary-200 dark:bg-secondary-700 hover:bg-secondary-300 hover:dark:bg-secondary-500 ml-2" %>
--- a/app/views/documents/new.html.erb
+++ b/app/views/documents/new.html.erb
@@ -37,7 +37,7 @@
                 :first, :last,
                 { selected: default_chunking_method },
                 { class: "rounded-md dark:bg-secondary-925" }
-               ) %>
+              ) %>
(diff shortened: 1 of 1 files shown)


nickthecook:
Hey, what exactly is being fixed here?

mattlindsey:
Submit button doesn't work (at least for me) I think because the f.submit line is outside for the begin/end for the form_for.  This fixes that.

nickthecook:
Ah, okay. They both work for me, so I'll merge this.

app/views/documents/new.html.erb
```diff
@@ -51,7 +50,8 @@
               <div class="flex-1"></div>
               <%= cp.text_field :overlap, value: default_chunk_overlap(default_chunking_method), size: 10, maxlength: 10, class: "rounded-md dark:bg-secondary-925 text-right" %>
             </div>
-          <% end %>
+            <% end%>
```
nickthecook, line 53:
```suggestion
            <% end %>
```

nickthecook:
Odd, the `f.submit` is not outside the form, just outside the `f.fields_for`. Any idea why this changes behaviour for you? Anything in your JS console when the button doesn't work?

app/views/documents/new.html.erb
nickthecook: Looks like some indentation might need an update too.

mattlindsey: In the original code the <% end %> doesn't match up with the form_for. I think rubocop and/or standard.rb can check erb files.  Rubocop does find some other minor fixes currently too.

mattlindsey:
> Odd, the `f.submit` is not outside the form, just outside the `f.fields_for`. Any idea why this changes behaviour for you? Anything in your JS console when the button doesn't work?

The <% end %> doesn't match up the the form_for in the original code.  I think rubocop or rubocop with 'standard' gem will check erb files if you want.  Also running 'rubocop' shows a bunch of other minor fixes, but nothing major.

nickthecook:
Looks like you've moved all the inputs into the rounded border.

The rounded border is intended to be around to upload field and link field, so it was clear the "OR" only applies to those two fields.

mattlindsey:
> Looks like you've moved all the inputs into the rounded border.
> 
> The rounded border is intended to be around to upload field and link field, so it was clear the "OR" only applies to those two fields.

I don't think so.  The border is the same

> Looks like you've moved all the inputs into the rounded border.
> 
> The rounded border is intended to be around to upload field and link field, so it was clear the "OR" only applies to those two fields.

Ah. I see.  The problem was just the <div> and fields_for were reversed, and that only caused a problem on my browser for some odd reason.

PR #93, opened 2024-10-06
Improve html parsing

mattlindsey:
I think this improves parsing a little by not calling Nokogiri::HTML multiple times when a file is parsed and chunked.
So let me know if this looks OK.
I'm pretty sure this works for all file types, but a few more rspec tests might be a good idea.

Also, I think we might not be splitting HTML recursively (like the Baran gem does for text).  I'm not exactly sure what 'recursively' should mean for HTML.
Here is how Langchain does it, so we can probably do it the same way: https://js.langchain.com/docs/how_to/code_splitter/#html

### DIFF
--- a/app/services/parsers/html.rb
+++ b/app/services/parsers/html.rb
@@ -1,7 +1,14 @@
 module Parsers
   class Html < Text
+    attr_reader :doc
+
+    def initialize(document)
+      super(document)
+      @doc = Nokogiri::HTML(@document.contents)
+    end
+
     def get_title
-      Nokogiri::HTML(@document.contents).css('title').text
+      @doc.css('title').text
     end
 
     private
--- a/app/services/parsers/html.rb
+++ b/app/services/parsers/html.rb
@@ -7,7 +7,7 @@ module Parsers
       @doc = Nokogiri::HTML(@document.contents)
     end
 
-    def get_title
+    def title
       @doc.css('title').text
     end
 
--- a/app/services/parsers/text.rb
+++ b/app/services/parsers/text.rb
@@ -18,7 +18,7 @@ module Parsers
     end
 
     # Overriede to set approriate title
-    def get_title
+    def title
       ''
     end
 
--- a/app/services/parsers/text.rb
+++ b/app/services/parsers/text.rb
@@ -9,7 +9,7 @@ module Parsers
 
     # Enumerable chunk records
     def chunks
-      chonker.chunk(@document)
+      chonker.chunk(self)
     end
 
     # Override to return the document's text if any additional processing is needed
--- a/app/services/parsers/text.rb
+++ b/app/services/parsers/text.rb
@@ -9,7 +9,7 @@ module Parsers
 
     # Enumerable chunk records
     def chunks
-      chonker.chunk(text)
+      chonker.chunk(@document)
     end
 
     # Override to return the document's text if any additional processing is needed
--- a/spec/factories/documents.rb
+++ b/spec/factories/documents.rb
@@ -9,3 +9,16 @@ FactoryBot.define do
     chunking_profile { association(:chunking_profile) }
   end
 end
+
+FactoryBot.define do
+  factory :documentx, class: "Document" do
+    chunks { [] }
+    user { association(:user) }
+    collection { association(:collection) }
+    filename { "gnu_manifesto.md" }
+    state { :chunked }
+    vector_id { nil } # TODO: we probably don't need this
+    chunking_profile { association(:chunking_profile) }
+    file { Rails.root.join("spec/fixtures/files/small_doc.md") }
+  end
+end
--- a/spec/support/shared_examples/services/chunkers/all_chunkers.rb
+++ b/spec/support/shared_examples/services/chunkers/all_chunkers.rb
@@ -1,16 +1,14 @@
 RSpec.shared_examples "all chunkers" do
   let(:document) { create(:documentx) }
+  let(:parser) { Parsers::CommonMark.new(document) }
+  let(:chunks) { subject.chunk(parser) }
 
   describe "#chunk" do
     it "returns Enumerable" do
-      parser = Parsers::CommonMark.new(document)
-      chunks = subject.chunk(parser)
       expect(chunks).to be_a(Enumerable)
     end
 
     it "returns chunk records" do
-      parser = Parsers::CommonMark.new(document)
-      chunks = subject.chunk(parser)
       expect(chunks.first).to be_a(Chunkers::ChunkRecord)
       expect(chunks.last).to be_a(Chunkers::ChunkRecord)
     end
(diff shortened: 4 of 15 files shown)


spec/support/shared_examples/services/chunkers/all_chunkers.rb
```diff
@@ -1,18 +1,16 @@
 RSpec.shared_examples "all chunkers" do
-  # Requires
-  # - `subject` to be instance of a chunker
-  # - the following `let` variables to be defined
-  #   - :text
-  #   - :text_type
```
nickthecook, line 6: Could this comment still be helpful?

spec/support/shared_examples/services/chunkers/all_chunkers.rb
```diff
@@ -1,18 +1,16 @@
...
+  let(:document) { create(:documentx) }
 
   describe "#chunk" do
-    let(:chunks) { subject.chunk(text) }
-
     it "returns Enumerable" do
+      parser = Parsers::CommonMark.new(document)
+      chunks = subject.chunk(parser)
```
nickthecook, line 7: The convention in RSpec is not to define test objects right in examples, but to do them beforehand with `let`. This would remove the duplication between this example and the next.

spec/support/shared_examples/services/chunkers/chunkers_with_overlap.rb
```diff
@@ -1,17 +1,11 @@
 RSpec.shared_examples "chunkers supporting overlap" do
-  # Requires
-  # - `subject` to be instance of a chunker
-  # - the following `let` variables to be defined
-  #   - :text
-  #   - :text_type
-  #   - :chunk_size
-  #   - :chunk_overlap
```
nickthecook, line 8: Not useful/accurate any longer?

spec/support/shared_examples/services/chunkers/chunkers_with_overlap.rb
```diff
@@ -1,17 +1,11 @@
...
+  let(:document) { create(:documentx) }
 
   describe "#chunk" do
-    let(:chunks) { subject.chunk(text) }
-
     context "with overlap" do
       it "returns chunks no bigger than 'chunk_size + chunk_overlap'" do
+        parser = Parsers::CommonMark.new(document)
+        chunks = subject.chunk(parser)
```
nickthecook, line 8: Same duplication.

spec/support/shared_examples/services/chunkers/chunkers_with_overlap.rb
```diff
@@ -21,6 +15,8 @@
       let(:chunk_overlap) { 0 }
 
       it "returns chunks no bigger than 'chunk_size'" do
+        parser = Parsers::CommonMark.new(document)
+        chunks = subject.chunk(parser)
```
nickthecook, line 19: Same.

app/services/parsers/html.rb
```diff
@@ -1,7 +1,14 @@
 module Parsers
   class Html < Text
+    attr_reader :doc
+
+    def initialize(document)
+      super(document)
+      @doc = Nokogiri::HTML(@document.contents)
+    end
+
     def get_title
```
nickthecook, line 10: Could this method be called just `title`?

app/services/parsers/text.rb
```diff
@@ -9,7 +9,7 @@ def initialize(document)
 
     # Enumerable chunk records
     def chunks
-      chonker.chunk(text)
+      chonker.chunk(self)
```
nickthecook, line 12: This seems odd, but anything else I can think of would introduce too much complexity to be worth it.

Once we support a couple more filetypes, the chunker-parser classes may need a bit of a refactor.

spec/factories/documents.rb
```diff
@@ -9,3 +9,16 @@
...
 end
+
+FactoryBot.define do
+  factory :documentx, class: "Document" do
+    chunks { [] }
+    user { association(:user) }
+    collection { association(:collection) }
+    filename { "gnu_manifesto.md" }
+    state { :chunked }
+    vector_id { nil } # TODO: we probably don't need this
```
nickthecook, line 20: You mean we probably don't need the field defined in the factory?

nickthecook (commented):
Nice, you got rid of the double-parsing of HTML docs.

nickthecook:
> Also, I think we might not be splitting HTML recursively (like the Baran gem does for text). I'm not exactly sure what 'recursively' should mean for HTML.

Yeah, I removed all but `p` from the list of HTML tags to use for splitting, because I was getting a bunch of very short chunks, from `h1`, `h2`, etc. The short chunks were an insanely low distance from some queries and they didn't provide any meaningful into, so I removed them.

One idea @oxaroky02 had was, for document types where we get headings, use those headings to contextualize chunks. E.g. when we feed the chunk to the LLM, it has text at the start of it that describes the section of a doc in which the chunk appears. This will help with various types of retrieval. I'm thinking the headings are useful for that, but not so much in the actual chunk body.

So I'm not sure what recursive means for HTML either - it's not like you're going to split on `p`, have a chunk that's too long, but then find a heading to split on within that `p`, right?

spec/support/shared_examples/services/chunkers/all_chunkers.rb
mattlindsey: `:text` and `:text_type` aren't required, but it does make me think we should be testing chunking all document types.  I think we're only testing .md files currently.

spec/support/shared_examples/services/chunkers/all_chunkers.rb
mattlindsey: fixed

spec/support/shared_examples/services/chunkers/chunkers_with_overlap.rb
mattlindsey: yes

spec/support/shared_examples/services/chunkers/chunkers_with_overlap.rb
mattlindsey: fixed

spec/support/shared_examples/services/chunkers/chunkers_with_overlap.rb
mattlindsey: fixed

app/services/parsers/html.rb
mattlindsey: fixed****

app/services/parsers/text.rb
mattlindsey: Yes.  Even having 2 different classes may not be useful/necessary.

spec/factories/documents.rb
nickthecook: Ah, you just copied and pasted it. :)

nickthecook:
Thanks, @mattlindsey !

PR #98, opened 2024-10-07
Handle url fetch error

mattlindsey:
Fixes problem where the site blows up hitting `View` on an imported URL that has errored, like on an invalid URL.

Also, I think the document error_message is supposed to show when you hover over `error`, but that doesn't appear to work.  Is that correct?

### DIFF
--- a/app/views/documents/_info.html.erb
+++ b/app/views/documents/_info.html.erb
@@ -5,7 +5,7 @@
   <div class="px-3">Chunk size: <strong><%= @document.chunking_profile&.size %></strong></div>
   <div class="px-3">Chunk overlap: <strong><%= @document.chunking_profile&.overlap %></strong></div>
   <div class="px-3 text-tertiary-700 dark:text-tertiary-600 hover:underline">
-    <%= link_to "Download", rails_blob_path(@document.file.blob, disposition: 'attachment') %>
+    <%= link_to "Download", rails_blob_path(@document.file.blob, disposition: 'attachment') if @document.file.present? %>
   </div>
   <!-- TODO: add page count, chunking profile, document size, download link -->
 </div>


app/services/fetch_web_document.rb
```diff
@@ -28,7 +28,17 @@ def set_document_attachment
...
 
   def fetch_file_content
-    content = HTTParty.get(@document.link)
+    begin
+      content = HTTParty.get(@document.link)
+    rescue StandardError => e
+      tempfile.write(e.message)
+      tempfile.rewind
+      set_document_attachment
+      tempfile.close
```
oxaronick, line 37: We should probably not put the error in the document. I think setting `error_message` is the right thing to do.

oxaronick:
Hover text is working for me:

image

oxaronick:
Can you rebase on `main`? I'm trying to set up CI with lint and unit tests, but it didn't run on your PR. I've put a fix in and I'm hoping you can try it.

oxaronick:
Are you using Chrome? Does the `title` attribute of `div`s show up for you on other pages?

mattlindsey:
Ahh. I just have to wait a couple of seconds for the error to appear on hover.

app/services/fetch_web_document.rb
```diff
@@ -28,7 +28,15 @@ def set_document_attachment
   end
 
   def fetch_file_content
-    content = HTTParty.get(@document.link)
+    begin
+      content = HTTParty.get(@document.link)
+    rescue StandardError => e
+      set_document_attachment
+      tempfile.close
+      @document.update!(state: :errored, error_message: e.message)
```
oxaronick, line 36: FetchWebDocumentJob should be doing this when the job retries are exhausted. Are you finding that that doesn't work?

Handling the errors in the job class is a bit of a pattern, because the job knows when it's giving up, whereas the service run by the job doesn't.

I don't even think you need to close the tempfile in the error case, since Tempfile will close the file description when its last reference is garbage collected.

app/services/fetch_web_document.rb
oxaronick: My mistake - the Job is doing this. See my other comment.

app/services/fetch_web_document.rb
mattlindsey: I think that since there's a definite error, it needs to error out here.  And if you don't close out the file it doesn't get attached and bombs out if anyone tries to view it.    
Maybe the `rescue` can be in the job, but you'd still want a blank file I think.  Otherwise you'd have to check for blank everywhere.

mattlindsey:
I think it works now by making sure fetch_file_content always attaches a file, blank or not. The error doesn't get set until after any retries.

oxaronick:
> I think it works now by making sure fetch_file_content always attaches a file, blank or not.

Maybe I'm just missing something, but why do we want to attach a file if we weren't able to download one?

> The error doesn't get set until after any retries.

That was the intention, since the user can at least wait for retries to run out before doing anything, like removing the document and trying again. If we set the status to `errored` while we're still working on it, they have no way of knowing we're going to try again or if we've given up.

oxaronick:
Ah, there is an error on documents#show if the file doesn't exist.

Try this in `app/views/documents/info.html.erb` instead of the `link_to` line that is there:

`<%= link_to "Download", rails_blob_path(@document.file.blob, disposition: 'attachment') if @document.file.present? %>`

app/services/fetch_web_document.rb
nickthecook: Since the file fetching moved to a job, there will always be a time when a document can have a nil `file`, so some checking is going to have to happen, but you're right that we don't want to do it everywhere.

Maybe a method in Document itself would be the place to do this. It would be better not to have all of its collaborators doing `document.file.blob` and such anyway.

PR #103, opened 2024-10-08
Fix HTML parsing / chunking by using markdown conversion (for now)

oxaroky02:
As a quick fix, I'm using `pandoc` to convert HTML to markdown for chunking.

### DIFF
--- a/app/services/parsers/html_via_markdown.rb
+++ b/app/services/parsers/html_via_markdown.rb
@@ -23,8 +23,8 @@ module Parsers
         @text = r[(r.index(ENDFRONTMATTER) + ENDFRONTMATTER.length)..]
       else
         error = e&.lines&.first || 'Unknown error running pandoc'
-        Rails.logger.error("Error running '#{cmd}' on HTML: #{@document.filename}\n#{error}")
-        raise StandardError, "Error converting HTML to MD: #{@document.filename}: #{error}'"
+        Rails.logger.error("Error running '#{CMD}' on HTML: #{@document.filename}\n#{error}")
+        raise StandardError, "Error converting HTML to MD: #{@document.filename}: #{error}"
       end
     end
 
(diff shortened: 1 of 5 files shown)


app/services/parsers/html_via_markdown.rb
```diff
@@ -0,0 +1,34 @@
...
+      grep -v '<!-- -->' |\
+      sed -e ':again' -e N -e '$!b again' -e 's/{[^}]*}//g'
+    TEXT
+
+    ENDFRONTMATTER = "---\n\n"
+
+    def initialize(document)
+      super(document)
+      # Convert the HTML into markdown, strip out all the junk we don't need
+      rawtxt, serr, status = Open3.capture3(CMD, stdin_data: @document.contents, binmode: true)
```
nickthecook, line 18: Smart, saving that extra character in the variable name. ;)

nickthecook, line 18:
```suggestion
      r, e, s = Open3.capture3(CMD, stdin_data: @document.contents, binmode: true)
```

app/services/parsers/html_via_markdown.rb
```diff
@@ -0,0 +1,34 @@
...
+      if status.success?
+        # extract the front matter from the raw text
+        @title = rawtxt.match(/title: (.+)\n/).captures.join
+        # remove the front matter to get the body
+        @text = rawtxt[(rawtxt.index("---\n\n") + 5)..]
+      else
+        Rails.logger.error("Error running '#{cmd}' on HTML: #{@document.filename}\n#{serr}")
+        raise StandardError, "Error converting HTML to markdown: #{@document.filename}'"
```
nickthecook, line 26: I imagine we don't want the full `stderr` here, but is there a predictable format to the errors such that you could add the first line of stderr and get something descriptive?

nickthecook: Actually, this should probably `raise` so the job fails.

oxaroky02: Fixed.

PR #104, opened 2024-10-10
Add jpg parsing

mattlindsey:
WIP. Comments?  Maybe the chunk could be compressed(?) image returned during similarity search?


nickthecook:
I'm not sure what to do with an image as a Document. I've vaguely assumed Archyve would support extracting images _from_ documents, perhaps storing them as an image Chunk, or an ImageChunk, or something like that. Then Archyve could generate a text description of the image and use that as the chunk's `embedding_content`. When the vector DB matches the text description during a similarity search, Archyve could do...  something that makes sense with it, like return it to an API client or display it for the user in chat. This is how I'd do it if the user uploaded a document that might contain images and text.

It looks like this approach is treating the image as the document, though, and I'm not sure how one could meaningfully chunk an image. Were you thinking of the image as a Document with one Chunk?

Do you think it makes more sense to treat an image as a Document, or a Chunk? Does it make sense to allow the user to upload images, or just documents that contain them?

Apologies if I'm missing something obvious!

mattlindsey:
Hi. It's fine. I can do it however you want, but might be useful to have image documents and when you add it, archive generates a the summary text by submitting to the llm and asking for it to provide one. That's a little different than what you're describing, so whatever you want is fine. Maybe nei4j already supports that - I dont know much about it.

nickthecook:
Yeah, could be useful to have image documents. I was thinking that they could be implemented as a special case of documents that contain images, each as a document with one chunk.

PDFs or other docs containing images would be good to support too, and I want to make sure we don't end up implementing things twice.

In terms of implementation of the latter, I expect images from a larger doc will each be a Chunk, so maybe for an image doc, it's one Document with one Chunk? Does that make sense?

If so, I think images like JPEGs would each be a Document, because that's the thing that has a file and belongs to a Collection, and it would have one chunk, because that's the thing that gets embedded and returned from a search.

If so (and let me know if you disagree) this PR would create a single chunk. There would be a job queued up after that could generate a text description, embed it, etc. Does that sound right?

Thanks for working in this, btw! Another cool feature.

mattlindsey:
Yeah.  A Document with a single image chunk sounds good.  I wanted to think about making sure it fits with pdf files too, so I created this issue #105, although I could have just added the comment to #3.

spec/services/parsers/jpg_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  let(:chunking_profile) { create(:chunking_profile, method: :basic, size: 800) }
+  let(:file) { fixture_file_upload("avatar.jpg", 'application/image') }
+  let(:document) { create(:document, state: :created, file:, chunking_profile:) }
+
+  context "when method is :basic" do
+    it_behaves_like "all parsers"
+
+    it "succeeds basic chunking" do
+      expect(subject.chunks.count).to eq(17)
```
mattlindsey, line 12: I'm having trouble getting this to succeed.  The site seems to work fine and creates right number of chunks for this file based on size, but fails here.  I added a 'chunk count' text to other classes and they succeed, but not here.  Kinda like `fixture_file_upload` isn't working for jpg.

oxaroky02:
Hola @mattlindsey, I'm jumping in to help Nick. If you don't mind I'm going to playback what I think you've committed so far so so I can get caught up.

- under `parsers/` 
  - adding a `Base` parser and deriving `Text` from that so we can start supporting non-text parsers. 
  - adding `PdfViaText` to replace `Pdf` so that we could support other ways to handle Pdfs in the future
  - adding `Image` parser as a base for image files
  - adding `Jpg` parser which is storing the binary data as Base64 encoded string
 - to `Chunkers`
   - adding `#self.single_chunker` which returns one `ChunkRecord` with the document's `#text` (which in the case of a JPG is the base64 encoded binary data)
   - adding a `:basic_image` chunking profile which I assume we would have to specify when ingesting an image

Anything I've missed that's significant?

mattlindsey:
Hey @oxaroky02. That's great! Feel free to jump in however you want. You can add code, comments or reviews and I can reply. That looks like a good playback. I think I saw that ollama takes images in base64, so the image chunk should be a useful feature.

oxaroky02:
> That looks like a good playback. I think I saw that ollama takes images in base64, so the image chunk should be a useful feature.

@mattlindsey, since ingesting an image means that the chunk will have the base64-encoding of the binary image data as the "text" and the "embedding", this PR is _a step towards_ properly ingesting an image and using a vision model (when we have that available in Archyve) to extract meaning from the image.  

Does that match your expectation of this PR's outcome? This PR is getting the ball rolling on image ingest.

mattlindsey:
@oxaroky02 Yeah. I think the code creates the single chunk but could you try it? It might need work. 
Another benefit we mentioned is returning images in a search of the documents(based on a description/summary) which might be a good step after this pr. Maybe hard code descriptions in rspec test fixtures and get that to work? After that maybe add them to the augmented context sent to the image capable llm?

mattlindsey:
I'm not totally clear on how the data should be stored in the chunks table, so what I did may not be ideal.

mattlindsey:
hi @oxaroky02.  I think I'm properly setting the data in the chunk table now, so if you choose a jpg file and 'basic_image' chunking on the Add Document page, it gets ingested into the single chunk.

If this looks ok so far, I think the next step can be submitting it to an llm to get a description to replace 'Unkown Image' (which what is embedded... so you can search on the word 'image' to see it on the Collection page).  We should also show the image itself on that page instead of the base64 data.

oxaroky02:
> If this looks ok so far, I think the next step can be submitting it to an llm to get a description to replace 'Unkown Image' 

@mattlindsey, I am going to work on a separate PR to add a `vision_model` setting and related internal changes to be able to make a request to a model with an image and query.  I'll ping you as soon as that is done and then you can update this PR.

oxaroky02:
@mattlindsey, PR 111 has been merged; you should now be able to

- locally, add a llava model config with vision=true via the admin portal,
- make it model for vision via the 'Use for vision' action in the admin portal, and
- use `Setting.vision_model` during parsing to get a model and use that with the LLM client API helper to ask for a description via the `#image` method. 

See `spec/lib/llm_clients/ollama/request_helper_spec.rb` line 94 for how to encode the image when calling the model client's #image method.

Let me know if you run into any issues. I

mattlindsey:
@oxaroky02 thanks. Out sick. Might take a bit.

PR #106, opened 2024-10-12
Enhance chat prompt augmentation

mattlindsey:
This seems to help, but ok if not deemed an improvement.  I need to fix tests and cleanup if this looks good.

### DIFF
--- a/app/models/chunk.rb
+++ b/app/models/chunk.rb
@@ -10,10 +10,6 @@ class Chunk < ApplicationRecord
   scope :embedded, -> { where.not(vector_id: nil) }
   scope :extracted, ->  { where(entities_extracted: true) }
 
-  # TODO: Take out next 2 lines if not needed.
-  # delegate :link, to: :document
-  # delegate :filename, to: :document
-
   def previous(count = 1)
     self.class.where(document:).where("id < ?", id).order(id: :asc).last(count)
   end
--- a/app/models/chunk.rb
+++ b/app/models/chunk.rb
@@ -10,6 +10,10 @@ class Chunk < ApplicationRecord
   scope :embedded, -> { where.not(vector_id: nil) }
   scope :extracted, ->  { where(entities_extracted: true) }
 
+  # TODO: Take out next 2 lines if not needed.
+  # delegate :link, to: :document
+  # delegate :filename, to: :document
+
   def previous(count = 1)
     self.class.where(document:).where("id < ?", id).order(id: :asc).last(count)
   end
--- a/app/services/prompt_augmentor.rb
+++ b/app/services/prompt_augmentor.rb
@@ -12,15 +12,22 @@ class PromptAugmentor
 
   def prompt
     @prompt ||= if @search_hits.any?
-      prompt = "You are given a question to answer based on some given textual context, all inside xml tags.\nIf the answer is not in the context but you think you know the answer, explain that to the user then answer with your own knowledge.\n\n" # rubocop:disable Layout/LineLength
+      # prompt = "You are given a query to answer based on some given textual context, all inside xml tags.\nIf the answer is not in the context but you think you know the answer, explain that to the user then answer with your own knowledge.\n\n" # rubocop:disable Layout/LineLength
+      prompt = <<~CONTENT
+        You are given a query to answer based on some given textual context, all inside xml tags.
+        If the answer is not in the context but you think you know the answer, explain that to the user then answer with your own knowledge.
+
+      CONTENT
       @search_hits.each do |hit|
         prompt << prompt_context(hit)
       end
-      prompt << "<user_question>\n#{@message.content}\n<user_question>\n"
+      prompt << "<user_query>\n#{@message.content}\n<user_query>\n"
     else
       @message.content
     end
   end
+
+  private
 
   def prompt_context(hit)
     if hit.document.web?
@@ -29,8 +36,6 @@ class PromptAugmentor
       "<context>\n<filename>#{hit.document.filename}</filename>\n<text>#{hit.content}</text>\n</context>\n"
     end
   end
-
-  private
 
   def link_message_with_augmentations
     @search_hits.each do |hit|
--- a/app/services/prompt_augmentor.rb
+++ b/app/services/prompt_augmentor.rb
@@ -29,6 +29,8 @@ class PromptAugmentor
   private
 
   def prompt_context(hit)
+    return "" unless hit.document
+
     if hit.document.web?
       "<context>\n<url>#{hit.document.link}</url>\n<scraped>#{hit.document.created_at}</scraped>\n<text>#{hit.content}</text>\n</context>\n" # rubocop:disable Layout/LineLength
     else
--- a/app/services/prompt_augmentor.rb
+++ b/app/services/prompt_augmentor.rb
@@ -10,7 +10,9 @@ class PromptAugmentor
     link_message_with_augmentations
   end
 
-  def prompt # rubocop:todo Metrics
+  def prompt
+    return nil if @search_hits.none?(&:relevant)
+
     @prompt ||= if @search_hits.any?
       prompt = <<~CONTENT
         You are given a query to answer based on some given textual context, all inside xml tags.
@@ -20,14 +22,11 @@ class PromptAugmentor
       CONTENT
 
       @search_hits.each do |hit|
-        if hit.document.nil?
-          prompt << "<context_item name=\"#{hit.name}\">\n<text>#{hit.content}</text>\n</context_item>\n" if hit.document.nil? # rubocop:todo Layout/LineLength
-        elsif hit.document.web?
-          prompt << "<context_item name=\"#{hit.name}\">\n<url>#{hit.document.link}</url>\n<scraped>#{hit.document.created_at}</scraped>\n<text>#{hit.content}</text>\n</context_item>\n" # rubocop:todo Layout/LineLength
-        else
-          prompt << "<context_item name=\"#{hit.name}\">\n<filename>#{hit.document.filename}</filename>\n<text>#{hit.content}</text>\n</context_item>\n" # rubocop:todo Layout/LineLength
-        end
+        prompt << "<context_item name=\"#{hit.name}\">\n"
+        prompt << context_item(hit)
+        prompt << "</context_item>\n"
       end
+
       prompt << "</context>\n\n"
       prompt << "Query: #{@message.content}\n"
     else
@@ -36,6 +35,16 @@ class PromptAugmentor
   end
 
   private
+
+  def context_item(hit)
+    if hit.document.nil?
+      "<text>#{hit.content}</text>\n"
+    elsif hit.document.web?
+      "<url>#{hit.document.link}</url>\n<scraped>#{hit.document.created_at}</scraped>\n<text>#{hit.content}</text>\n"
+    else
+      "<filename>#{hit.document.filename}</filename>\n<text>#{hit.content}</text>\n"
+    end
+  end
 
   def link_message_with_augmentations
     @search_hits.each do |hit|
--- a/app/services/prompt_augmentor.rb
+++ b/app/services/prompt_augmentor.rb
@@ -11,34 +11,33 @@ class PromptAugmentor
   end
 
   def prompt
-    if @search_hits.none?(&:relevant)
-      return "The query found hits, but none were relevant. Query: #{@message.content}\n"
+    if @search_hits.none?
+      return nil
+    elsif @search_hits.none?(&:relevant)
+      return "The query found hits, but none were relevant.\nQuery: #{@message.content}\n"
     end
 
-    @prompt ||= if @search_hits.any?
-      prompt = <<~CONTENT
-        You are given a query to answer based on some given textual context, all inside xml tags.
-        If the answer is not in the context but you think you know the answer, explain that to the user then answer with your own knowledge.
+    prompt = <<~CONTENT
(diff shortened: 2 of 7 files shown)


spec/factories/chunks.rb
```diff
@@ -4,5 +4,9 @@
     vector_id { SecureRandom.uuid }
     content { Faker::Lorem.sentence(word_count: 10) }
     embedding_content { Faker::Lorem.sentence(word_count: 16) }
+
+    factory :chunk_from_web do
```
nickthecook, line 8: Does this still work if you move it out of the `factory :chunk do` block?

app/services/prompt_augmentor.rb
```diff
@@ -12,12 +12,16 @@ def augment
 
   def prompt
     @prompt ||= if @search_hits.any?
-      prompt = "Here is some context that may help you answer the following question:\n\n"
+      prompt = "You are given a question to answer based on some given textual context, all inside xml tags.\nIf the answer is not in the context but you think you know the answer, explain that to the user then answer with your own knowledge.\n\n"
```
nickthecook, line 15: I like it! I put very little thought into that prompt. This looks better.

spec/factories/chunks.rb
mattlindsey: Yup. Looks like I just add `parent: :chunk`.
I'll fix that, and the tests soon.

app/services/prompt_augmentor.rb
```diff
@@ -12,21 +12,24 @@ def augment
 
   def prompt
     @prompt ||= if @search_hits.any?
-      prompt = "You are given a question to answer based on some given textual context, all inside xml tags.\nIf the answer is not in the context but you think you know the answer, explain that to the user then answer with your own knowledge.\n\n"
+      prompt = "You are given a question to answer based on some given textual context, all inside xml tags.\nIf the answer is not in the context but you think you know the answer, explain that to the user then answer with your own knowledge.\n\n" # rubocop:disable Layout/LineLength
```
nickthecook, line 15: Converting to a HEREDOC would let you avoid the `rubocop:disable`.

spec/services/prompt_augmentor_spec.rb
```diff
@@ -6,32 +6,34 @@
...
+      # Search::SearchHit.new(chunks[1], 220.0),
     ]
   end
 
+  # TODO: make part of #prompt test below
+  describe "#prompt for doc from web" do
+    it "has the correct content" do
+      c = create(:chunk_from_web)
+      hs = [Search::SearchHit.new(c, 240.0)]
+      p = described_class.new(message, hs)
```
nickthecook, line 24: Can you convert these into `let`s?

spec/services/prompt_augmentor_spec.rb
```diff
@@ -6,32 +6,34 @@
...
   let(:search_hits) do
     [
       Search::SearchHit.new(chunks[0], 200.0),
-      Search::SearchHit.new(chunks[1], 220.0),
+      # Search::SearchHit.new(chunks[1], 220.0),
     ]
   end
 
+  # TODO: make part of #prompt test below
+  describe "#prompt for doc from web" do
```
nickthecook, line 20: Most of the RSpec examples in Archyve (and in general, I think) follow this structure:

```ruby
describe "#prompt" do  # string indicates the instance method `prompt` is being tested by this block
  let(:some_var) { "some_val" }  # something you need to calculate or run before each test, to keep the test clean and/or reuse code
  let(:some_other_var) { "some_other_val" }

  it "has the correct content" do
    expect(subject.prompt).to eq(some_var)
  end
end
```

nickthecook: Oh, there is a section for `#prompt` - you can add these examples there!

spec/services/prompt_augmentor_spec.rb
```diff
@@ -6,32 +6,34 @@
...
+    it "has the correct content" do
+      c = create(:chunk_from_web)
+      hs = [Search::SearchHit.new(c, 240.0)]
+      p = described_class.new(message, hs)
+      # TODO: Figure out why needed:
+      allow_any_instance_of(Document).to receive(:web?).and_return(true) # rubocop:disable RSpec/AnyInstance
```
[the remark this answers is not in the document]
nickthecook: Oh, I see the `web?` call you added now, but I guess that still doesn't explain why you need to do this.

What was the error you were getting without this line?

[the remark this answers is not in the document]
mattlindsey: Without that line the generated prompt has wrong elements (i.e., `<filename>` instead of `<url>` and `<scraped>`).  I think `web?` returns true instead of false because the `document_from_web` and `chunk_from_web` factories I added aren't quite working correctly.

app/services/prompt_augmentor.rb
mattlindsey: I might do that, but I'm finding it easier to do this way for now.

spec/services/prompt_augmentor_spec.rb
mattlindsey: Yeah. I was planning removing the lets and putting it in `#prompt` but  I added this test temporarily to try to figure out why I can't get it to work without the allow_any_instance_of below.

spec/services/prompt_augmentor_spec.rb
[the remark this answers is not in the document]
mattlindsey: It works great if I don't use inheritance, and it may be a little easier to understand, so I think I'll do this:

```ruby
FactoryBot.define do
  factory :chunk do
    document { association(:document) }
    vector_id { SecureRandom.uuid }
    content { Faker::Lorem.sentence(word_count: 10) }
    embedding_content { Faker::Lorem.sentence(word_count: 16) }
  end
  factory :chunk_from_web, class: 'Chunk' do
    document { association(:document_from_web) }
    vector_id { SecureRandom.uuid }
    content { Faker::Lorem.sentence(word_count: 10) }
    embedding_content { Faker::Lorem.sentence(word_count: 16) }
  end
end

**Comment by mattlindsey** [comment]
I think this PR looks ok now.
There a few lines in chunk.rb and search_hit.rb that may help searching, but I'm not sure exactly how that works.  I can remove or add them if desired.

File: app/models/chunk.rb [hunk 7]
```diff
@@ -10,6 +10,10 @@ class Chunk < ApplicationRecord
   scope :embedded, -> { where.not(vector_id: nil) }
   scope :extracted, ->  { where(entities_extracted: true) }
 
+  # TODO: Take out next 2 lines if not needed.
+  # delegate :link, to: :document
+  # delegate :filename, to: :document
+
```
nickthecook, line 16: My rule of thumb is not to add speculative stuff, even if it looks like it might be handy. So, might as well take them out?

app/services/search/search_hit.rb
```diff
@@ -34,5 +34,20 @@ def distance_increase_ratio
...
+    #   @reference.link if @reference.respond_to?(:link)
+    # end
+
+    # def created_at
+    #   @reference.created_at if @reference.respond_to?(:created_at)
+    # end
+
+    # def filename
+    #   @referemce.filename if @reference.respond_to?(:filename)
+    # end
```
nickthecook, line 51: I guess these were not needed either?

spec/factories/documents.rb
```diff
@@ -3,13 +3,29 @@
...
+
     trait :with_file do
       file { Rails.root.join("spec/fixtures/files/small_doc.md") }
     end
   end
+  factory :document_from_web, class: 'Document' do
+    chunks { [] }
+    user { association(:user) }
+    collection { association(:collection) }
+    state { :chunked }
```
nickthecook, line 20: Not a big deal at all, but if this is in state `chunked`, it would normally have chunks. What if the factory assumed an empty set of chunks and state `created`?

app/services/prompt_augmentor.rb
```diff
@@ -12,19 +12,30 @@ def augment
...
+      prompt << "<user_query>\n#{@message.content}\n<user_query>\n"
     else
       @message.content
     end
   end
 
   private
 
+  def prompt_context(hit)
+    if hit.document.web?
```
nickthecook, line 32: In my local env I'm getting a NoMethodError from `.web?` here:

```
jobs    | NoMethodError: undefined method `web?' for nil:NilClass
jobs    |
jobs    |     if hit.document.web?
jobs    |                    ^^^^^/Users/nick/src/github.com/nickthecook/archyve-main/app/services/prompt_augmentor.rb:32:in `prompt_context'
```

Upon investigation, it looks like it's because sometimes a SearchHit is a Chunk, and has a `.document`, but sometimes a SearchHit is a GraphEntity and doesn't. In PromptAugmentor, you're going to have to handle both of these cases. In the future, there may be even more cases, like `hit.reference` being an instance of GraphRelationship.

Do you have any Collections with knowledge graph enabled?

mattlindsey: Oh. I see. It's a little confusing, but I fixed to check that `hit.document` is not nil.   Maybe there should be rspec tests for the different things that can be referenced? (which would have failed earlier)

mattlindsey: But now I'm not sure if that's correct.  If there's a hit on a GraphEntity, shouldn't I return something instead of blank context?
```ruby
return "" unless hit.document

Comment (nickthecook) [reply app/services/prompt_augmentor.rb:34]: Yeah, you found augmentations.

If you want to feed the link to the LLM, put it in the prompt context. If you just want it available for the user, put the info in the augmentation.

There are partials for displaying augmentations differently depending on whether they're chunks or entities, so you could just update the partial for displaying chunks and add the extra info there.

You could also do one PR with just the new prompt format and the XML tags, and do the link stuff in a separate PR.

Comment (mattlindsey) [reply app/services/prompt_augmentor.rb]: I added a test for the search hit on an entity.  I'm sure this can be cleaned up, but maybe would be nice to try it and see how it performs?

Also, there's an error on test github step.

Comment (nickthecook) [reply app/services/prompt_augmentor.rb]: So, what do you think? Can you just add the prompt change without the link info showing up in the UI for now, and add the augmentation stuff later? Seems like most of the complications in this one are around getting the link from the hit, whereas your prompt improvement is clear and simple.

Comment (mattlindsey) [reply app/services/prompt_augmentor.rb]: So the link gets added to the context now and I think everything works and the tests pass, but you want to wait to do that better later?

Comment (mattlindsey) [reply app/services/prompt_augmentor.rb]: Oh wait.  Sorry.  Now I see the error in `rspec ./spec/services/graph/summarize_collection_entities_spec.rb`.

I'm hoping it's minor and will fix it tonight or tomorrow.

Comment (mattlindsey) [reply app/services/prompt_augmentor.rb]: Fixed.  Sorry about that.

Comment (nickthecook) [reply app/services/prompt_augmentor.rb]: No worries!

I was just thinking that maybe feeding the link to the LLM isn't going to help, but I'll have a look and can adjust later.

File: app/services/prompt_augmentor.rb [hunk 11]
```diff
@@ -10,14 +10,26 @@ def augment
     link_message_with_augmentations
   end
 
-  def prompt
+  def prompt # rubocop:todo Metrics
```
nickthecook, line 13: Can you address the Rubocop exemptions? There are many in the codebase now, but I'm trying hard not to add more.

There are four in this file and this one is pretty broad, and I worry the code will end up littered with them if we're not more disciplined.

mattlindsey: Addressed rubocop in app code, but rspec was driving me a little crazy since I couldn't set `relevant` in the search hit object. I tried `Search::SearchHit.new(graph_entity, 240.0, relevant: true)` but that didn't work.

I added a check to avoid updating the message prompt with `Query: ...` when there are no relevant hits, which was confusing.  
Maybe it should add `The query found hits, but none were relevant` in that case?

nickthecook:
I think most of the churn on this PR has been around including the link info or document name in the context, which I was really hoping you'd split out into a separate PR. Just doing the prompt improvement, with the XML-style tags, was an easy win.

Do you think including the document metadata in the context improves the quality of answers? If you do, I'll merge this and fix some of the other stuff up.

app/services/prompt_augmentor.rb
```diff
@@ -11,22 +11,46 @@ def augment
...
 
   def prompt
-    @prompt ||= if @search_hits.any?
-      prompt = "Here is some context that may help you answer the following question:\n\n"
-      @search_hits.each do |hit|
-        prompt << "#{hit.content}\n\n"
-      end
+    if @search_hits.none?
+      return nil
+    elsif @search_hits.none?(&:relevant)
```
nickthecook, line 16: Are you seeing search hits that aren't relevant returned?

In a chat scenario, you shouldn't be, and V1::ChatController doesn't allow the client to specify `include_irrelevant`.

mattlindsey: I think I was just seeing empty(none?) search hits and unnecessarily augmented prompt in the ui.

mattlindsey:
It seems to help the llm respond more helpfully, and I can a simple new pr tomorrow or Monday if you want. Or you can fix this if you want. Maybe just the 'relevant' stuff should be removed?

nickthecook:
> It seems to help the llm respond more helpfully, and I can a simple new pr tomorrow or Monday if you want. Or you can fix this if you want. Maybe just the 'relevant' stuff should be removed?

Yeah, if you remove the check for relevant hits and just check if there are hits or not, I'll merge this!

PR #110, opened 2024-10-19
Add xml context and logic to augmented prompt

mattlindsey:
This seems to improve llm responses a little bit.

This PR replaces earlier one that got out of hand! My confusion was that the current ui can indicate that the prompt is being augment by showing `> augmentations` and there are none when you click the arrow.  I fixed that too, which I hope is ok.

### DIFF
--- a/app/services/prompt_augmentor.rb
+++ b/app/services/prompt_augmentor.rb
@@ -11,19 +11,39 @@ class PromptAugmentor
   end
 
   def prompt
-    @prompt ||= if @search_hits.any?
-      prompt = "Here is some context that may help you answer the following question:\n\n"
-      @search_hits.each do |hit|
-        prompt << "#{hit.content}\n\n"
-      end
-
-      prompt << "Question: #{@message.content}\n"
-    else
-      @message.content
+    if @search_hits.none?
+      return nil
     end
+
+    prompt = <<~CONTENT
+      You are given a query to answer based on some given textual context, all inside xml tags.
+      If the answer is not in the context but you think you know the answer, explain that to the user then answer with your own knowledge.
+
+      <context>
+    CONTENT
+
+    @search_hits.each do |hit|
+      prompt << "<context_item name=\"#{hit.name}\">\n"
+      prompt << context_item_content(hit)
+      prompt << "</context_item>\n"
+    end
+
+    prompt << "</context>\n\n"
+    prompt << "Query: #{@message.content}\n"
+    prompt
   end
 
   private
+
+  def context_item_content(hit)
+    if hit.document.nil?
+      "<text>#{hit.content}</text>\n"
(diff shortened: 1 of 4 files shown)


nickthecook:
Hey @mattlindsey, sorry for the delay in getting back to this.

I did some testing with [this article](https://en.wikipedia.org/wiki/Hurricane_Irene_(2005)) on Hurricane Irene. I asked llama3.1 "How strong was Irene", and I got this reply:

> Based on the provided textual context, I can answer your query directly.
>
> According to the text within the XML tags:
> 
>  "Irene peaked as a Category 2 hurricane" in Chunk 117.
>  "Initially a powerful hurricane in the Atlantic Ocean during its peak strength phase before landfall," in Irene (event).
> 
> However, for your question about how strong Irene was, I can provide more information from my external knowledge.
> 
> Hurricane Irene in 2005 had a maximum sustained wind speed of 105 mph (165 km/h), which corresponds to a Category 2 hurricane on the Saffir-Simpson Hurricane Wind Scale.

Here's another example, and excerpt from a response, where it now seems very concerned with being precise, probably to a point that isn't useful:

> Based on the provided textual context, I can answer your query. However, since you're asking for "more" information, I'll provide a detailed explanation of my process and then share some additional knowledge that complements the context.

I like the idea of having the LLM aware that it's providing an answer based on information from some system, and not just chatting. I would like to have the system "aware" that its job is to summarize and select information for a user from a database. However, this response is quite wordy and talks directly about a lot of technical details that I don't think a user needs if they're trying to interrogate their documents. "Chunks" and "XML tags" aren't things a user should need to understand. We probably need to "speak" to the model in terms that we want it to use with the user.

I can see updating the prompt or using a system prompt to explain to the model that it is playing the role of a librarian, and the user may be asking about documents. I think if we gave the model info on what document a piece of text came from, and it told the user the info came from that document, that might be good. The model could even say things like "I'm sorry, I couldn't find that in the information I have", and that would be an interaction that anyone could understand.

I'm not sure we're quite ready for this though - @oxaroky02 is planning on doing some work to contextualize chunks, so information like section name or page are available, and then we could have the model give more a useful reference to the documents. I'd be happy to chat more about where I hope we can take Archyve in the future. I'll start keeping an eye on Discord in case you want to take me up on that.

If the move to XML tags helps, I'm all for that, and if there's a way to have the model be clear about what it found or didn't based on what's in the vector DB, without becoming too wordy or nerdy, that would be good too.

Thoughts?

mattlindsey:
Yeah. When I was trying it with the default phi model I don't think it mentioned XML tags or such. Hopefully this PR gave some good feedback and I don't mind if you close it unless someone wants to continue it. An enhancement would be to save the xml as a modifiable 'template' but that's kinda elaborate and may not help a lot.

PR #134, opened 2024-11-05
Refactor Chunk properties, and update all related code to reflect changes

oxaroky02:
The crux of this PR is as follows:

In `Chunk` and `ChunkRecord`
- rename `content` to `excerpt` - this will be the document's excerpt represented by the chunk
- add `surrounding_content` - this will be the version of the excerpt with any additional prefix and suffix text if there's any overlap criteria used by the chunker
- add `headings` - the location of the excerpt based on previous headings
- add `location_summary` - some text that describes the chunk's "location" within the document.
- retain `embedding_content` - this will be used to store the input (calculated based on above properties) to the embeddings that is used in the similarity search

All the other changes are a result of making the code compatible with the above changes. 

Not all the new properties are in use yet; the current changes preserve backwards compatibility with any previously ingested content.

### DIFF
--- a/app/models/chunk.rb
+++ b/app/models/chunk.rb
@@ -10,6 +10,13 @@ class Chunk < ApplicationRecord
   scope :embedded, -> { where.not(vector_id: nil) }
   scope :extracted, ->  { where(entities_extracted: true) }
 
+  # Require embedding content when creating new chunks
+  # Schema doesn't enforce presence for backwards compatibility with existing data
+  validate :has_explicit_embedding_content?, on: :create
+
+  # Ensure that someone can't change a previously set embedding content back to `nil`
+  attr_readonly :embedding_content
+
   def previous(count = 1)
     self.class.where(document:).where("id < ?", id).order(id: :asc).last(count)
   end
@@ -21,6 +28,14 @@ class Chunk < ApplicationRecord
   def embedding_content
     # This will help with any previously ingested documents which will
     # have a nil for this column
-    self[:embedding_content] || content
+    self[:embedding_content] || excerpt
+  end
+
+  private
+
+  def has_explicit_embedding_content?
+    return if self[:embedding_content]
+
+    errors.add(:embedding_content, "New chunks require embedding content")
   end
 end
--- a/app/models/chunk.rb
+++ b/app/models/chunk.rb
@@ -16,6 +16,7 @@ class Chunk < ApplicationRecord
 
   # Ensure that someone can't change a previously set embedding content back to `nil`
   attr_readonly :embedding_content
+  attr_readonly :excerpt, :surrounding_content, :location_summary, :headings
 
   def previous(count = 1)
     self.class.where(document:).where("id < ?", id).order(id: :asc).last(count)
--- a/app/models/chunk.rb
+++ b/app/models/chunk.rb
@@ -12,7 +12,7 @@ class Chunk < ApplicationRecord
 
   # Require embedding content when creating new chunks
   # Schema doesn't enforce presence for backwards compatibility with existing data
-  validate :has_explicit_embedding_content?, on: :create
+  validate :explicit_embedding_content?, on: :create
 
   # Ensure that someone can't change a previously set embedding content back to `nil`
   attr_readonly :embedding_content
@@ -33,8 +33,8 @@ class Chunk < ApplicationRecord
 
   private
 
-  def has_explicit_embedding_content?
-    return if self[:embedding_content]
+  def explicit_embedding_content?
+    return false if self[:embedding_content]
 
     errors.add(:embedding_content, "New chunks require embedding content")
   end
--- a/app/models/chunk.rb
+++ b/app/models/chunk.rb
@@ -14,10 +14,6 @@ class Chunk < ApplicationRecord
   # Schema doesn't enforce presence for backwards compatibility with existing data
   validate :explicit_embedding_content?, on: :create
 
-  # Ensure that someone can't change a previously set embedding content back to `nil`
-  attr_readonly :embedding_content
-  attr_readonly :excerpt, :surrounding_content, :location_summary, :headings
-
   def previous(count = 1)
     self.class.where(document:).where("id < ?", id).order(id: :asc).last(count)
   end
--- a/app/services/chunk_document.rb
+++ b/app/services/chunk_document.rb
@@ -11,7 +11,12 @@ class ChunkDocument
     parser.chunks.each do |chunk_record|
       chunk = Chunk.create!(
         document: @document,
-        content: chunk_record.content,
+
+        excerpt: chunk_record.excerpt,
+        headings: chunk_record.headings,
+        location_summary: chunk_record.location_summary,
+        surrounding_content: chunk_record.surrounding_content,
+        embedding_content: chunk_record.embedding_content,
         embedding_content: chunk_record.embedding_content)
 
       EmbedChunkJob.perform_async(chunk.id)
--- a/app/services/chunk_document.rb
+++ b/app/services/chunk_document.rb
@@ -7,7 +7,14 @@ class ChunkDocument
     reset_document unless @document.created?
 
     @document.chunking!
+    create_chunk_records
+    @document.update(title: parser.title)
+    @document.chunked!
+  end
 
+  private
+
+  def create_chunk_records
     parser.chunks.each do |chunk_record|
       chunk = Chunk.create!(
         document: @document,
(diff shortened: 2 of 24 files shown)


app/services/chunkers/chunk_record.rb
```diff
@@ -3,16 +3,27 @@ module Chunkers
...
+      # vector embeddings
+      text = ""
+      if headings
+        text << headings << "\n"
+      end
+      if location_summary
+        text << location_summary << "\n"
+      end
+      # Use document excerpt if no surrounding content provided (compatibility)
+      text << (surrounding_content || excerpt) << "\n"
```
oxaronick, line 26: Most of the time Archyve is using two newlines to separate semantically different blocks, the same way one would separate paragraphs in a text or markdown doc.

I'm not sure how much it helps the LLM, but: I think it's better to be consistent, it reads better to a user, and one would think it matches training data better.

spec/models/chunk_spec.rb
```diff
@@ -1,60 +1,82 @@
...
+  let(:embedding_content) { "surround subject sound" }
   let(:chunks) { document.chunks.sort }
 
-  before do
-    create(:chunk, document:)
-    create(:chunk, document:)
-    document.chunks << subject
-    create(:chunk, document:)
-    create(:chunk, document:)
+  describe "creating a chunk" do
```
oxaronick, line 11: Not a big deal, but if you're looking for an arg to `describe` that's consistent with the pattern of naming after methods, you could use `create!`.

spec/models/chunk_spec.rb
```diff
@@ -1,60 +1,82 @@
...
+    describe "#embedding_content" do
+      # context "when embedding_content is not set" do
+      #   before do
+      #     chunk = subject
+      #     chunk.embedding_content = nil
+      #   end
+
+      #   it "returns the excerpt" do
+      #     expect(subject.embedding_content).to eq(excerpt)
+      #   end
```
oxaronick, line 75: Ddi you meant to leave this commented out?

app/models/chunk.rb
```diff
@@ -10,6 +10,13 @@ class Chunk < ApplicationRecord
   scope :embedded, -> { where.not(vector_id: nil) }
   scope :extracted, ->  { where(entities_extracted: true) }
 
+  # Require embedding content when creating new chunks
+  # Schema doesn't enforce presence for backwards compatibility with existing data
+  validate :explicit_embedding_content?, on: :create
+
+  # Ensure that someone can't change a previously set embedding content back to `nil`
+  attr_readonly :embedding_content
```
oxaronick, line 18: I was wondering if `excerpt` should be read-only as well, since arguably it is even less valid to change a field that is meant to be a direct excerpt from a document.

Then I started wondering if all these new fields should be read-only - if you can't change the `embedding_content` then you shouldn't be changing the `headers` etc.

Then I wondered if it was worth enforcing this at all, or if we should just trust our code to only update these if it makes sense, like if we're reprocessing a bunch of chunks in place with a new algorithm. :)

Thoughts?

oxaronick (commented):
Looking good!

app/services/chunk_document.rb
```diff
@@ -7,20 +7,27 @@ def execute
...
 
     @document.chunking!
+    create_chunk_records
+    @document.update(title: parser.title)
+    @document.chunked!
+  end
 
+  private
+
+  def create_chunk_records
```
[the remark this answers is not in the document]
oxaroky02: Oh, right. Yes.

app/services/chunkers/chunk_record.rb
oxaroky02: Yes, makes sense. The expectation should be that the fields themselves aren't required to (nor should they) include terminating newlines, and then the aggregator will use two new lines between them.

spec/models/chunk_spec.rb
oxaroky02: Yes, thanks.

spec/models/chunk_spec.rb
oxaroky02: Ah, good catch. I need to rewrite that using a double since creating with `embedding_content: nil` raises an error now, but there might be chunks in the DB where that column is null.

app/models/chunk.rb
oxaroky02: Right. And yes, I think it is best if these were read-only as well. While today many will have `nil` values, over time they will only have values when created during ingesttion. Even a re-ingestion will create new chunks.

oxaronick: OK. So is the resolution to make all these attributes read-only, or to remove the read-only property from `embedding_content`?

Or did you already make one of those changes and I'm not seeing it in GitHub for some reason?

oxaronick: OK... I just looked down.

Minor, but why is `embedding_content` on its own line?

spec/models/chunk_spec.rb
```diff
@@ -1,60 +1,106 @@
...
+        expect(subject.next(3)).to eq(chunks[3..4])
+      end
     end
-  end
 
-  describe "#next" do
-    it "returns the next chunk" do
-      expect(subject.next).to contain_exactly(chunks[-2])
+    describe "#embedding_content=" do
+      it "fails since property is readonly" do
```
oxaronick, line 61: This is a good candidate for a `shared_example` now.

spec/models/chunk_spec.rb
oxaroky02: For context, after offline discussion, reverted the use of `attr_readonly`. It's an internal thing anyway, and if we need protections better to do it differently.

spec/models/chunk_spec.rb
oxaroky02: N/a since these are gone now. 😄

app/models/chunk.rb
oxaroky02: Readonly removed as discussed offline.

oxaroky02:
@oxaronick, all the e2e tests passed as well.
```
ops test-e2e --tag llm
Running `TEST_E2E=true bundle exec rspec --tag "~skip" spec/e2e --tag llm` in environment 'dev'...
Run options:
  include {:llm=>true}
  exclude {:skip=>true}
 12/12 |====================================================================================================== 100 =======================================================================================================>| Time: 00:00:51

Finished in 51.78 seconds (files took 0.4816 seconds to load)
12 examples, 0 failures
 rcgtky02@KOOKA  archyve-fork   main ●  dev  ops test-e2e --tag opp
Running `TEST_E2E=true bundle exec rspec --tag "~skip" spec/e2e --tag opp` in environment 'dev'...
Run options:
  include {:opp=>true}
  exclude {:skip=>true}
 11/11 |====================================================================================================== 100 =======================================================================================================>| Time: 00:00:24

Finished in 24.49 seconds (files took 0.54059 seconds to load)
11 examples, 0 failures
 rcgtky02@KOOKA  archyve-fork   main ●  dev  ops test-e2e
Running `TEST_E2E=true bundle exec rspec --tag "~skip" spec/e2e` in environment 'dev'...
Run options: exclude {:skip=>true}
 50/50 |====================================================================================================== 100 =======================================================================================================>| Time: 00:00:36

Finished in 36.36 seconds (files took 0.50839 seconds to load)
50 examples, 0 failures
 rcgtky02@KOOKA  archyve-fork   refactor-chunk-properties  dev  ops test-e2e
Running `TEST_E2E=true bundle exec rspec --tag "~skip" spec/e2e` in environment 'dev'...
Run options: exclude {:skip=>true}
 50/50 |====================================================================================================== 100 =======================================================================================================>| Time: 00:00:53

Finished in 53.11 seconds (files took 0.4932 seconds to load)
50 examples, 0 failures
```

oxaronick:
LGTM!

mattlindsey:
@oxaroky02 Hi. Quick question since you're working on chunks.  
Can you think of how I should determine if a chunk is an image/jpg so that I know to submit it to the vision llm to get a description of it?

oxaroky02:
> Can you think of how I should determine if a chunk is an image/jpg so that I know to submit it to the vision llm to get a description of it?

Nick and I have been discussing this in fact, and I'm working on introducing a little bit of workflow to make this easier to implement. Right now you're treating an image as a base64 text chunk _first_, and I think we should flip that and have a prior step that loads the image and talks to the vision model _before_ chunking. 

Your current implementation clarified this for us. I have a draft PR coming shortly which should put something in place to make this easier.

PR #136, opened 2024-11-06
Define Mediator to help dispatch document ingest

oxaroky02:
Created `Mediator` service to handle the ingest document workflow. The name is a play on "media" since this service will handle taking a document through it's various stages depending on its media type.

For now
- if document has no filename
  - if document has a web link
    - dispatch a `FetchWebDocumentJob`
  - else raise error
- else
  - dispatch a `ChunkDocumentJob`

In addition, the `FetchWebDocumentJob` now calls the `Mediator.ingest` after it completes downloading the link. Currently, this triggers chunking, but over time we can handle more interim steps before chunking depending on the document's media type.

### DIFF
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -1,5 +1,46 @@
 require 'rails_helper'
 
 RSpec.describe Document do
-  pending "add some examples to (or delete) #{__FILE__}"
+  subject { described_class.new(state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
+
+  let(:collection) { create(:collection) }
+  let(:user) { create(:user) }
+  let(:file) { nil }
+  let(:filename) { nil }
+  let(:link) { nil }
+  let(:chunking_method) { :basic }
+  let(:chunking_profile) { create(:chunking_profile, method: chunking_method, size: 800) }
+
+  describe "when document has a web link" do
+    let(:link) { "https://en.wikipedia.org/wiki/Tabloid_journalism" }
+
+    it "#web? is true" do
+      expect(subject.web?).to be true
+    end
+
+    it "#file is not attached" do
+      expect(subject.file.attached?).to be false
+    end
+
+    it "#filename is nil" do
+      expect(subject.filename).to be_nil
+    end
+  end
+
+  describe "when document has a file" do
+    let(:file) { fixture_file_upload("gnu_manifesto.md", 'application/html') }
+    let(:filename) { "spec/fixtures/files/gnu_manifesto.md" }
+
+    it "#web? is false" do
+      expect(subject.web?).to be false
+    end
+
+    it "#file is attached" do
+      expect(subject.file.attached?).to be true
+    end
+
+    it "#filename is present" do
+      expect(subject.filename).not_to be_nil
+    end
+  end
 end
--- a/spec/services/mediator_spec.rb
+++ b/spec/services/mediator_spec.rb
@@ -14,30 +14,14 @@ RSpec.describe Mediator do
     context "with web link" do
       let(:link) { "https://en.wikipedia.org/wiki/Tabloid_journalism" }
 
-      it "fetches web page" do
-        # Catching the message seems to prevent the actual method from being called
-        # So limiting this to example
+      before do
         allow(FetchWebDocumentJob).to receive("perform_async")
+      end
 
+      it "fetches web page" do
         subject.ingest(doc)
 
         expect(FetchWebDocumentJob).to have_received("perform_async").with(doc.id)
-      end
-
-      context "when web page has been fetched" do
-        before do
-          allow(ChunkDocumentJob).to receive("perform_async")
-        end
-
-        it "chunks document" do
-          subject.ingest(doc)
-
-          # Need to drain to cause the job to be run so I can check if the
-          # chunking job was initiated after fetching
-          FetchWebDocumentJob.drain
-
-          expect(ChunkDocumentJob).to have_received("perform_async").with(doc.id)
-        end
       end
     end
 
--- a/spec/services/mediator_spec.rb
+++ b/spec/services/mediator_spec.rb
@@ -11,31 +11,23 @@ RSpec.describe Mediator do
   let(:doc) { Document.new(state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
 
   describe "#ingest document" do
-    after do
-      FetchWebDocumentJob.clear
-      ChunkDocumentJob.clear
-    end
-
     context "with web link" do
       let(:link) { "https://en.wikipedia.org/wiki/Tabloid_journalism" }
-      # let(:doc) { create(:document, state: :created, filename: nil, link: "https://en.wikipedia.org/wiki/Tabloid_journalism", chunking_profile:) }
 
       it "fetches web page" do
         expect do
           subject.ingest(doc)
         end.to change(FetchWebDocumentJob.jobs, :size).by(1)
+        Sidekiq::Worker.clear_all
       end
 
       context "when web page has been fetched" do
-        before do
-          allow(Document).to receive(:find).and_return(doc)
-        end
-
         it "chunks document" do
-          expect do
-            subject.ingest(doc)
-            FetchWebDocumentJob.drain
-          end.to change(ChunkDocumentJob.jobs, :size).by(1)
+          doc.save
+          subject.ingest(doc)
+
+          expect { FetchWebDocumentJob.drain }.to change(ChunkDocumentJob.jobs, :size).by(1)
+          Sidekiq::Worker.clear_all
         end
       end
     end
@@ -43,12 +35,12 @@ RSpec.describe Mediator do
     context "with filename" do
       let(:file) { fixture_file_upload("gnu_manifesto.md", 'application/html') }
       let(:filename) { "spec/fixtures/files/gnu_manifesto.md" }
-      # let(:doc) { create(:document, state: :created, filename: "spec/fixtures/files/gnu_manifesto.md", chunking_profile:) }
 
       it "chunks document" do
         expect do
           subject.ingest(doc)
         end.to change(ChunkDocumentJob.jobs, :size).by(1)
+        Sidekiq::Worker.clear_all
       end
     end
   end
--- a/spec/services/mediator_spec.rb
+++ b/spec/services/mediator_spec.rb
@@ -8,26 +8,35 @@ RSpec.describe Mediator do
   let(:link) { nil }
   let(:chunking_method) { :basic }
   let(:chunking_profile) { create(:chunking_profile, method: chunking_method, size: 800) }
-  let(:doc) { Document.new(state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
+  let(:doc) { create(:document, state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
 
   describe "#ingest document" do
     context "with web link" do
       let(:link) { "https://en.wikipedia.org/wiki/Tabloid_journalism" }
 
       it "fetches web page" do
-        expect do
-          subject.ingest(doc)
-        end.to change(FetchWebDocumentJob.jobs, :size).by(1)
-        Sidekiq::Worker.clear_all
+        # Catching the message seems to prevent the actual method from being called
+        # So limiting this to example
+        allow(FetchWebDocumentJob).to receive("perform_async")
+
+        subject.ingest(doc)
+
+        expect(FetchWebDocumentJob).to have_received("perform_async").with(doc.id)
       end
 
       context "when web page has been fetched" do
+        before do
+          allow(ChunkDocumentJob).to receive("perform_async")
+        end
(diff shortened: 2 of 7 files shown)


oxaronick:
Makes sense so far.

I would have gone for a `class`, but there's really no need for that, since it's just a method with some logic in it, and probably will stay that way. Nice.

spec/services/mediator_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  describe "#ingest document" do
+    after do
+      FetchWebDocumentJob.clear
+      ChunkDocumentJob.clear
+    end
+
+    context "with web link" do
+      let(:link) { "https://en.wikipedia.org/wiki/Tabloid_journalism" }
+      # let(:doc) { create(:document, state: :created, filename: nil, link: "https://en.wikipedia.org/wiki/Tabloid_journalism", chunking_profile:) }
```
nickthecook, line 21: Can this (and the commented line below) be removed?

nickthecook:
Normally I'd test a Job by asserting that it was called, or by calling `perform` and checking its behaviour. Here, it looks like you're calling your class and checking Sidekiq state to see that the right things happen. I don't see a big difference between these two, but a side effect is that half your spec is managing Sidekiq state.

Does the Sidekiq RSpec stuff not do all the draining and clearing automatically? If not, this seems like we're at higher risk of having tests interfere with each other because we forgot to call `clear` or `drain` or something like that.

I can still be convinced, but right now it seems to me that this way of testing Jobs (which, in the end, are just classes with a public `perform` method that can be easily called from a spec the old-fashioned way) is more complex and requires us to manage Sidekiq state ourselves.

spec/services/mediator_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      context "when web page has been fetched" do
+        before do
+          allow(Document).to receive(:find).and_return(doc)
+        end
+
+        it "chunks document" do
+          expect do
+            subject.ingest(doc)
+            FetchWebDocumentJob.drain
```
nickthecook, line 37: What is this doing? Does it need to be in the `expect` block?

spec/services/mediator_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    context "with filename" do
+      let(:file) { fixture_file_upload("gnu_manifesto.md", 'application/html') }
+      let(:filename) { "spec/fixtures/files/gnu_manifesto.md" }
+      # let(:doc) { create(:document, state: :created, filename: "spec/fixtures/files/gnu_manifesto.md", chunking_profile:) }
+
+      it "chunks document" do
+        expect do
+          subject.ingest(doc)
+        end.to change(ChunkDocumentJob.jobs, :size).by(1)
```
nickthecook, line 51: Could this be an `expect {}` instead of an `expect do`? Seems like it would fix on one line, and read better.

nickthecook:
Looks good, just questions about the testing.

spec/services/mediator_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      it "fetches web page" do
+        expect do
+          subject.ingest(doc)
+        end.to change(FetchWebDocumentJob.jobs, :size).by(1)
+      end
+
+      context "when web page has been fetched" do
+        before do
+          allow(Document).to receive(:find).and_return(doc)
```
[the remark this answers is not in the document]
oxaroky02: Because the document created in this test has a `nil` id. The `FetchWebDocumentJob` called `Document.find` using the document ID and does not find it. 

Trying to answer your question made me realise that instead of this hack I could use `doc.save` instead.

spec/services/mediator_spec.rb
oxaroky02: The ingest call does not need to be in the `expect` block.

spec/services/mediator_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  let(:filename) { nil }
+  let(:link) { nil }
+  let(:chunking_method) { :basic }
+  let(:chunking_profile) { create(:chunking_profile, method: chunking_method, size: 800) }
+  let(:doc) { Document.new(state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
+
+  describe "#ingest document" do
+    after do
+      FetchWebDocumentJob.clear
+      ChunkDocumentJob.clear
```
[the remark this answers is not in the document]
oxaroky02: Makes sense. I've replace the `after` block with a single call to `Sidekiq::Worker.clear_all` at the end of each example, making a further assumption that no example will leave jobs behind for future use.

This doesn't address your earlier question about using job queue counting instead of more specific behaviour. I'm doing it this way because I don't want the `Mediator` tests to know what the jobs do per se, but only care that certain sequencing happens. For now Mediator is a very simple if/else. Soon after this PR however I plan to introduce a flow based on media types, and at that point the tests might not need to check for jobs.

oxaroky02:
> I can still be convinced, but right now it seems to me that this way of testing Jobs (which, in the end, are just classes with a public `perform` method that can be easily called from a spec the old-fashioned way) is more complex and requires us to manage Sidekiq state ourselves.

My reasoning is that I'm testing `Mediator#ingest` who just relies on initiating jobs based on some condition, and expecting some jobs to turn around and call `Mediator#ingest` again. On one hand I had use some knowledge of the jobs that get launched, on the other hand I really don't want to test the jobs per se. 

What I mean is, I need to call `#ingest` and test that works, rather than call `#perform` on a job ... at least, not in this spec. 

(TODO, the jobs are missing tests, so maybe the actual job behaviour tests would go there.)

oxaronick:
> What I mean is, I need to call `#ingest` and test that works, rather than call `#perform` on a job ... at least, not in this spec.

Agreed - you don't want to be testing the jobs, just that Mediator is enqueueing them with the right args.

spec/services/mediator_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  subject { described_class }
+
+  let(:collection) { create(:collection) }
+  let(:user) { create(:user) }
+  let(:file) { nil }
+  let(:filename) { nil }
+  let(:link) { nil }
+  let(:chunking_method) { :basic }
+  let(:chunking_profile) { create(:chunking_profile, method: chunking_method, size: 800) }
+  let(:doc) { Document.new(state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
```
oxaronick, line 11: Try `create(:document, ...)`. That will save it for you, too.

Or, if you would prefer not to use the factory, just use `Document.create!` here.

spec/services/mediator_spec.rb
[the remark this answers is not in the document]
oxaronick: I think if you tested that Mediator is making the right call to these job classes, you wouldn't need to check what's in the sidekiq queue at all.

E.g.:

```
before do
  allow(FetchWebDocumentJob).to receive("perform_async")`.
end

it "something something" do
  subject.ingest
  expect(FetchWebDocumentJob).to have_received("perform_async").with(doc.id)`.
end
```

1. This checks that your class enqueues the right job (as your spec does now)
2. The job doesn't actually run (as with your spec now)
3. This checks that it passed the right arg (which your spec does not do now)
4. We don't need to remember to call anything in an `after` block or risk introducing flakiness (not true with your spec right now)

There are examples of this in `spec/lib/sidekiq_cron_schedule_spec.rb`.

I won't mention this again - if you really want to merge this way, that's cool with me. At this point I'll leave it up to you.

spec/models/document_spec.rb
```diff
@@ -1,5 +1,46 @@
 require 'rails_helper'
 
 RSpec.describe Document do
-  pending "add some examples to (or delete) #{__FILE__}"
+  subject { described_class.new(state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
```
oxaronick, line 4: Good to have specs for this!

spec/services/mediator_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    context "with filename" do
+      let(:file) { fixture_file_upload("gnu_manifesto.md", 'application/html') }
+      let(:filename) { "spec/fixtures/files/gnu_manifesto.md" }
+
+      it "chunks document" do
+        expect do
+          subject.ingest(doc)
+        end.to change(ChunkDocumentJob.jobs, :size).by(1)
+        Sidekiq::Worker.clear_all
```
oxaronick, line 43: To me, duplicating it at the end of every example is worse than putting in an `after` block, because now we have the opportunity to forget it on every example, instead of just when creating a new spec!

The `after` block is perfectly dependable - I didn't mean to cast aspersions on it. What I was commenting on is that if we test job enqueueing this way, and we add a new spec, now we need to remember the `after` block or we introduce flakiness. Whereas testing jobs the way I suggested there is no risk of flakiness and no need to remember to clear a job queue in each spec.

spec/services/mediator_spec.rb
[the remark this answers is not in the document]
oxaroky02: I agree, that is much better than checking the jobs list. Thanks for the example, let me try.

spec/services/mediator_spec.rb
oxaroky02: Earlier on I had a lot of trouble with `create(:document ...)` but now, I guess since I'm passing all parameters from the test, when I switched it works. Sheesh.

spec/services/mediator_spec.rb
oxaroky02: Ah, I misunderstood. Still getting used to the conventions.

Re sidekiq jobs, from what I can tell, the rspect support by default queues up jobs in the job class. It does mean that un-drained job will accumulate, and possibly across examples. I ran into issues with the job counts. So makes sense if I can avoid relying on those in the first place.

spec/services/mediator_spec.rb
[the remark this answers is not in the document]
oxaroky02: @oxaronick, take a look now. I was able to get rid of the job counting by using strategy above. 

I had to keep one use of `#drain` to test the flow through the fetch job to the chunk job. Does that make sense? From what I can tell, using Sidekiq in test queues the job but until I call `#drain` the job won't run.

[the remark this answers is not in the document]
oxaronick: That makes sense - you don't actually want the job to run, do you?

Sidekiq should be in test mode just based on the `$RAILS_ENV`, so it should happily queue your jobs up and then die with a full queue at the end of the test run.

spec/services/mediator_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        before do
+          allow(ChunkDocumentJob).to receive("perform_async")
+        end
+
+        it "chunks document" do
+          subject.ingest(doc)
+
+          # Need to drain to cause the job to be run so I can check if the
+          # chunking job was initiated after fetching
+          FetchWebDocumentJob.drain
```
oxaronick, line 37: Looks like this is testing FetchWebDocumentJob. This makes sense, but should be in that class' spec, not here.

If you avoid testing the other class in this spec, you can throw your `allow` into a `before` block, and this spec will get nice and clean.

oxaronick: Oh, did you do this because you want to pass control to the job, and then have that job pass it back to this class?

It's better to avoid depending on another class' behaviour in a spec (with some common exceptions like ActiveRecord classes). This class should have certain behaviour, and if getting called by FetchWebDocumentJob is one of those, just make sure you have a context for it in this spec. I wouldn't rely on the other class to call a second instance of the class you're testing.

spec/services/mediator_spec.rb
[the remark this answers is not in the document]
oxaroky02: I only want the jobs to run _when_ I need to test the sequencing. For e.g. `FetchWebDocumentJob` needs to run to then call `Mediator#ingest` to continue to chunking.

spec/services/mediator_spec.rb
oxaroky02: In the case where `doc` is just a web link we have a flow like this:
- `Mediator#ingest`
- `FetchWebDocumentJob#perform`
- `Mediator#ingest`
- `ChunkDocumentJob#perform`

Normally I would have tested fetch web doc behaviour in it's own test, but in this case it's calling `Mediator#ingest` ... so I put this here. Regardless of where I put it, though, I need `#perform` to be executed to test the right job was invoked.

spec/services/mediator_spec.rb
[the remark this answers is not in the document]
oxaroky02: Per our chat, see latest commit where we have a simpler `Mediator` test where it has no insight into the implementation of `FetchWebDocumentJob`, but rather tests it's inputs.

spec/services/mediator_spec.rb
oxaroky02: I've added test spec for `FetchWebDocumentJob` to test it's behaviour separately from the `Mediator` spec.

oxaronick:
Nice!

PR #139, opened 2024-11-13
Document chaining and typing

oxaroky02:
This is a "steps along the way" PR with a bunch of small changes that get us ready to add content conversion in a later PR.

### Document

- Adding `parent` property `Document` to prepare for managing converted documents so we can track original document separately from the converted document which gets chunked
- Adding content type-related methods based on attached file's API (ActiveStorage support content type detection)

### Mediator

- Refactoring `#ingest` to better describe what it can do
  - `#next_step` returns verb (`:fetch`, `:convert`, `:chunk`) or `nil` if it doesn't know what to do
  - ingest is now a case statement that is easier to visualise
  - conversion is not implemented (and will not be in this PR)
- Defined base `IngestError` for all the `Mediator` errors.

### Parsers

- Adding utility `#textual?` since this is where we know what formats we can parse as "text"
- Modified how parsers are selected by
  - adding a `self.can_parse?` to each parser
  - going through a list of `ENABLED` parsers to pick one that can parse the document
- Include content type in determining suitable parser, which allows us to download and parse all supported formats

### FetchWebDocument

- Tweaked to store temp file in binary mode to support fetching non-text files
- Using URL's path's extension if any for the temp file name

### Controllers

- Fixed `DocumentsController#vectorize` to handle errors raised by `Mediator.ingest` and flash them to the user


app/services/mediator.rb
```diff
@@ -1,15 +1,28 @@
...
+  class DocumentNotChunkable < StandardError; end
+
+  # Returns true if the document's content type is chunkable
+  def self.chunkable?(document)
+    document.file.attached? && Parsers.textual?(document.content_type)
+  end
+
+  # Return true if document needs to be fetched?
+  def self.fetchable?(document)
+    document.filename.nil?
```
oxaronick, line 12: This doesn't necessarily make the Document fetchable. It's more that it requires fetching, or needs to be fetched, or doesn't have a file.

spec/models/document_spec.rb
```diff
@@ -10,8 +10,9 @@
   let(:link) { nil }
   let(:chunking_method) { :basic }
   let(:chunking_profile) { create(:chunking_profile, method: chunking_method, size: 800) }
+  let(:parent) { nil }
 
-  describe "when document has a web link" do
+  describe "with a web link" do
```
oxaronick, line 15: I might try refactoring this spec so we follow the `describe "#methodname"` convention, but for now you could just change all the `describe` calls to `context`. They read more like `context`s.

spec/models/document_spec.rb
```diff
@@ -22,25 +23,113 @@
...
+    end
+
+    it "has file name" do
       expect(subject.filename).not_to be_nil
     end
   end
+
+  describe "with a JPEG file" do
+    let(:filename) { "spec/fixtures/files/avatar.jpg" }
+    let(:file) { fixture_file_upload("avatar.jpg") }
```
oxaronick, line 58: [Avatar](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://www.youtube.com/watch%3Fv%3DjVhlJNJopOQ&ved=2ahUKEwi0-N21ityJAxXTLDQIHdUMJHkQtwJ6BAhQEAI&usg=AOvVaw07Kl3r7aQzWXfjx8m2_Dsi)

app/services/mediator.rb
oxaroky02: I didn't like this either; new impl coming with something more expressive

spec/models/document_spec.rb
oxaroky02: For now I'll switch to `context`.

oxaronick: Reading more of this spec, I feel like just using `context` will be fine as long as it's possible to come up with example names that read well and make the linter happy.

I sometimes run into an issue with RSpec where I want to describe some behaviour but I need to call multiple methods of the subject to describe the behaviour. The `describe "#methodname` pattern gets in the way of this. So I think for some classes - maybe all models - we could skip `describe` blocks and go straight to `context`.

app/services/mediator.rb
oxaroky02: Done.

oxaronick:
Looks pretty good!

app/controllers/documents_controller.rb
```diff
@@ -38,6 +38,9 @@ def destroy
 
   def vectorize
     Mediator.ingest(@document)
+  rescue Mediator::IngestError
+    flash[:error] = ["Unable to ingest document. Likely an unsupported format."]
```
oxaronick, line 42: Any reason you're not capturing the exception as `=> e` and rendering its message?

oxaroky02: Rushing. 😄  Fixed.

oxaroky02:
@mattlindsey FYI this PR _intentionally_ breaks the image ingest. As of now the "ingestion" pipeline is ready for me to put in "conversion" semantics. One more PR to go 🤞 and we can turn image -> text -> chunking on.

PR #140, opened 2024-11-15
Document conversion during ingest

oxaroky02:
This PR defines a conversion system that allows us to define content format converters. As an example the PDF parser (which was previously a chunker) has been changed to a converter that takes a PDF as input and produces a text document as output.  When an ingested document (A) requires conversion, we end up with a child document (B) whose content is then chunked. 

To make all this work
- Refactored Mediator into an instance-based service
- Tweaked Parsers to better reflect intent of methods (and make implementation consistent)
- Removed PDF from enabled parsers list
- Enabled PDF-to-text converter so that PDFs get converted into text documents which then get chunked
- Refactored document reset/deletion to work with documents with children
- Modified collections view to show child documents nested under the top-level ones
- Tests tests tests ... for my sanity

The ingest flow looks like this (high level):

```mermaid
graph LR
  User([User])

  User -- upload doc --> Fetch --> Convert --> Chunk
```

After this PR is merged, when a PDF is ingested the following will occur:

```mermaid
graph LR
  User([User])
  ChromaDB[(ChromaDB)]

  User -- upload PDF -->  Convert --create child text doc --> Chunk --> ChromaDB
  User -- link to online PDF --> Fetch -- create doc --> Convert

```


mattlindsey:
Hi.  Not necessary, but it would be cool if you or Nick could add a little blurb somewhere (maybe here: https://github.com/nickthecook/archyve?tab=readme-ov-file#document-ingest) describing the steps of ingesting a document and the 'philosophy' of how they're stored in Archyve.   There are a lot of steps and services in the code, so it's a little hard to understand.

oxaronick:
Documents that I already had ingested have lost their "Reprocess" buttons. 

image

Is this intentional?

oxaronick:
Ingesting a new doc from a URL resulted in this:

```ruby
jobs    | 2024-12-03T19:32:24.135Z pid=50287 tid=w37 WARN: NameError: uninitialized constant ChunkDocument::DocumentResetHelper
jobs    |
jobs    |     DocumentResetHelper.new(@document).execute
jobs    |     ^^^^^^^^^^^^^^^^^^^
jobs    | Did you mean?  DocumentsHelper
jobs    | 2024-12-03T19:32:24.136Z pid=50287 tid=w37 WARN: app/services/chunk_document.rb:35:in `reset_document'
jobs    | app/services/chunk_document.rb:7:in `execute'
jobs    | app/sidekiq/chunk_document_job.rb:14:in `perform'
```

Might need `Helpers::` in front of the class name?

oxaronick (commented):
First pass review; haven't looked at tests or UI yet.

oxaroky02:
> Documents that I already had ingested have lost their "Reprocess" buttons.

Actually :-) I moved the reprocess button next to the Delete button with an icon. I can change it back to a labeled button if you prefer; I meant to discuss with you but forgot.

spec/jobs/convert_document_job_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      it "calls Mediator after conversion" do
+        subject.perform(doc.id)
+
+        expect(Mediator).to have_received("ingest").with(subject.new_doc)
+      end
+
+      it "produces a new document" do
+        subject.perform(doc.id)
+
+        expect(subject.new_doc).not_to be_nil
```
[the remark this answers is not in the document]
oxaroky02: Makes sense.

app/services/converters/pdf_to_text.rb
```diff
@@ -0,0 +1,27 @@
+module Converters
+  class PdfToText < Base
+    def convert
+      # Don't repeat conversion; use a new Converter
+      return unless ready?
```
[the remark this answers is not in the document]
oxaroky02: Good point. It's my paranoia shining through. Removed.

app/services/converters/base.rb
```diff
@@ -0,0 +1,66 @@
...
+      tempfile.write(content)
+      tempfile.rewind
+      filename = File.basename(tempfile.path)
+      new_doc = Document.new(
+        filename:,
+        parent: input_document,
+        chunking_profile: input_document.chunking_profile,
+        collection: input_document.collection,
+        user: input_document.user)
+      new_doc.file.attach(io: tempfile, filename:)
```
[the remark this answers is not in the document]
oxaroky02: > [Tempfile.create](https://docs.ruby-lang.org/en/master/Tempfile.html#method-c-create) creates a usual File object. The timing of file deletion is predictable. Also, it supports open-and-unlink technique which removes the temporary file immediately after creation.

I've fixed to use `Tempfile.create do ... end` to ensure cleanup happens.

app/services/helpers/document_reset_helper.rb
```diff
@@ -0,0 +1,55 @@
+module Helpers
+  class DocumentResetHelper
```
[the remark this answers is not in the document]
oxaroky02: I moved "reset" logic into the helper, and `DestroyDocument` is the service that after a reset destroys the top-level document. I.e. `DocumentResetHelper` cleans up children and associated content but _never_ destroys the input document.

app/services/destroy_document.rb
```diff
@@ -0,0 +1,12 @@
+class DestroyDocument
```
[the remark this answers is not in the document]
oxaronick: Ah, I missed the enqueue of that job right at the beginning.

app/services/mediator.rb
```diff
@@ -1,40 +1,45 @@
...
+      end
     elsif document.file.attached?
-      if Parsers.textual?(document.content_type)
+      if Parsers.can_chunk?(document)
         :chunk
-      elsif need_conversion?(document)
+      elsif Converters.can_convert?(document)
         :convert
       end
     end
```
oxaronick, line 43: Would this class be simpler if these conditionals were directly in `#next_step`?

app/views/collections/_document.html.erb
```diff
@@ -11,27 +20,50 @@
...
+          <%= link_to "View", collection_document_path(document.collection, document), class: "rounded-lg bg-secondary-700 dark:bg-secondary-400 text-secondary-100 dark:text-secondary-900 py-1 px-4 disabled:bg-opacity-40 mr-2" %>
+          <% if document.stop_jobs %>
+            <%= button_to "Enable jobs", collection_document_start_path(document.collection, document), class: "rounded-lg px-4 py-1 text-center text-white align-baseline dark:text-secondary-950 block font-medium bg-secondary-700 dark:bg-secondary-400" %>
+          <% else %>
+            <%= button_to "Disable jobs", collection_document_stop_path(document.collection, document), class: "rounded-lg px-4 py-1 text-center text-white align-baseline dark:text-secondary-950 block font-medium bg-secondary-700 dark:bg-secondary-400" %>
+          <% end %>
+        </div>
+      <% end %>
+
+      <% unless document.no_children? %>
```
oxaronick, line 41: Not a big deal, but feels like a double-negative.

I try to name methods on objects in a positive way, like `has_children?` to avoid this kind of thing. You could even go with `childless?`.

app/services/mediator.rb
oxaroky02: I don't understand ... do you mean run the jobs without returning the symbols for the case/when?

spec/jobs/convert_document_job_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      let(:file) { fixture_file_upload("gnu_manifesto.pdf") }
+
+      before do
+        allow(Mediator).to receive("ingest")
+      end
+
+      it "calls Mediator after conversion" do
+        subject.perform(doc.id)
+
+        expect(Mediator).to have_received("ingest")
```
oxaronick, line 27: That's  lot of setup for just this example.

Should you also check that `ingest` got the right arg, using `.with(doc)` or something like that?

oxaronick: Looks like the code should also:

- update the doc after the conversion
- not call `Mediator#ingest` if the doc is `errored?`

Should those be tested as well?

spec/models/document_spec.rb
```diff
@@ -96,6 +96,23 @@
     end
   end
 
+  describe "#children" do
+    let(:parent) { create(:document, parent: nil) }
+
+    before do
+      parent.save
```
oxaronick, line 103: You're using `create`, so you shouldn't need to save it explicitly here.

spec/models/document_spec.rb
```diff
@@ -132,4 +149,21 @@
...
   end
+
+  describe "#destroy" do
+    context "when document has children" do
+      let(:grandparent) { create(:document) }
+      let(:parent) { create(:document, parent: grandparent) }
+
+      before do
+        grandparent.save
+        parent.save
```
oxaronick, line 160: Don't need these two saves.

It's quite rare to see `.save` in a before block in a spec. Usually you just create the object in the way you want:

- `create` if you want it saved (it's analogous to ActiveRecord's `create`)
- `build` or `described_class.new` if you don't

In the case of this spec, `subject` was created with `described_class.new` so a `subject.save` in only some `describe` blocks make sense. You shouldn't need to do it for anything else, though.

spec/services/converters_spec.rb
```diff
@@ -0,0 +1,32 @@
+RSpec.describe Converters do
+  subject { described_class }
```
oxaronick, line 2: Better to just use `described_class` throughout the spec instead of `subject`. Usually `subject` is an instance of `described_class`.

spec/services/destroy_document_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  let(:file) { fixture_file_upload("gnu_manifesto.md") }
+  let(:filename) { "spec/fixtures/files/gnu_manifesto.md" }
+  let(:link) { nil }
+  let(:chunking_method) { :basic }
+  let(:chunking_profile) { create(:chunking_profile, method: chunking_method, size: 800) }
+  let(:doc) { create(:document, state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
+
+  describe "#execute" do
+    before do
+      doc.save
```
oxaronick, line 15: Not needed.

oxaronick: I will stop mentioning these now.

app/views/collections/_document.html.erb
oxaroky02: I've replaced that one with `if document.children.any?` to avoid the double negative.

spec/services/mediator_spec.rb
```diff
@@ -25,7 +25,22 @@
...
+    context "with a PDF file" do
+      let(:filename) { "spec/fixtures/files/gnu_manifesto.pdf" }
+      let(:file) { fixture_file_upload("gnu_manifesto.pdf") }
+
+      before do
+        allow(ConvertDocumentJob).to receive("perform_async")
+      end
+
+      it "converts document" do
+        subject.ingest(doc)
```
oxaronick, line 37: I missed this last time: `use `described_class` when you're making a call to `Mediator`, and `subject` when you're making a call to an instance of `Meditator`.

spec/services/parsers_spec.rb
```diff
@@ -30,13 +30,13 @@
...
 
-    include_examples "supported_format_parser_by_filename", ".pdf", Parsers::Pdf
+    # include_examples "supported_format_parser_by_filename", ".pdf", Parsers::Pdf
     include_examples "supported_format_parser_by_filename", ".docx", Parsers::Docx
     include_examples "supported_format_parser_by_filename", ".md", Parsers::CommonMark
     include_examples "supported_format_parser_by_filename", ".txt", Parsers::Text
     include_examples "supported_format_parser_by_filename", ".html", Parsers::HtmlViaMarkdown
 
-    include_examples "supported_format_parser_by_content_type", "application/pdf", Parsers::Pdf
+    # include_examples "supported_format_parser_by_content_type", "application/pdf", Parsers::Pdf
```
oxaronick, line 39: Did you mean to leave these in?

spec/jobs/convert_document_job_spec.rb
oxaroky02: This is related to the earlier comment about `ConvertDocumentJob` exposing `new_doc` attribute. I would need to re-expose that to use `.with(...)` in this test. At this point, all I care about is that the conversion sends a new doc to `Mediator`. The converters are themselves tested on their own.

app/services/mediator.rb
oxaronick: Yes.

I may be missing something, but it looks like, of these two methods, one decides what to do and one does it. But, since "doing it" is just a one-liner, you could "do the thing" instead of telling the other method to do the thing.

spec/jobs/convert_document_job_spec.rb
oxaronick: You can use `Document.last` instead of `new_doc`, and add an example that checks that `Document.count` went up by 1, so you know you have the right doc.

spec/services/destroy_document_spec.rb
oxaroky02: When I remove the `doc.save` and `child.save` in these specs, I get test failures:

```
  1) DestroyDocument#execute with childless document destroys document
     Failure/Error: expect { subject.execute }.to change(Document, :count).by(-1)
       expected `Document.count` to have changed by -1, but was changed by 0
     # ./spec/services/destroy_document_spec.rb:24:in `block (4 levels) in <main>'


  2) DestroyDocument#execute when document has children destroys both documents document
     Failure/Error: expect { subject.execute }.to change(Document, :count).by(-2)
       expected `Document.count` to have changed by -2, but was changed by 0
     # ./spec/services/destroy_document_spec.rb:18:in `block (4 levels) in <main>'

 2/2 |============================================================================= 100 =============================================================================>| Time: 00:00:00

Finished in 0.59534 seconds (files took 3.88 seconds to load)
2 examples, 2 failures
```

app/services/mediator.rb
oxaroky02: Because of the error handling, the single method versions that also raise an ingest error all result in complexity warnings by rubocop.

spec/jobs/convert_document_job_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  subject { described_class.new }
+
+  let(:collection) { create(:collection) }
+  let(:user) { create(:user) }
+  let(:file) { nil }
+  let(:filename) { nil }
+  let(:link) { nil }
+  let(:chunking_method) { :basic }
+  let(:chunking_profile) { create(:chunking_profile, method: chunking_method, size: 800) }
+  let(:doc) { create(:document, state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
```
oxaronick, line 13:
```suggestion
  let!(:doc) { create(:document, state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
```

oxaronick: That might not help, but it might...

oxaronick: Wrong spec -  you're not having trouble with this one.

spec/services/destroy_document_spec.rb
oxaronick: What's happening is that `doc` isn't evaluated until you call `subject.execute` inside the block where you expect it to change the count. So, within the block:

- `doc` is created (+1)
- `child` is created (+1)
- `child` is destroyed (-1)
- `doc` is destroyed (-1)

for a total change of 0. :)

Since `child` will cause `doc` to be evaluated, you could change `let(:child)` to `let!(:child)`.

spec/services/reset_document_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  let(:chunking_profile) { create(:chunking_profile, method: chunking_method, size: 800) }
+  let(:doc) { create(:document, state: :created, link:, file:, filename:, chunking_profile:, collection:, user:) }
+
+  describe "#execute" do
+    before do
+      doc.save
+    end
+
+    context "when document has children" do
+      let(:child) { create(:document, parent: doc) }
```
oxaronick, line 19: Same here:

```suggestion
      let!(:child) { create(:document, parent: doc) }
```

oxaroky02:
> ... describing the steps of ingesting a document and the 'philosophy' of how they're stored in Archyve. There are a lot of steps and services in the code, so it's a little hard to understand.

@mattlindsey, see PR description for some high level diagrams. These will end up in `docs/` as well. After we merge this, to ingest images and convert them into textual representation for chunking,
- write a converter that takes `image/jpg` and produces `text/markdown`; see `app/services/converters/` for example.
- register it in the `ENABLED` array in `Converters` module

The content ingest flow would work like this:

```mermaid
graph LR
  User([User])
  ChromaDB[(ChromaDB)]
  Convert[Convert JPG <br/>to Text]
  style Convert fill:#f9f,stroke:#333,stroke-width:2px
  Ollama@{shape: odd, label: "AI<br/>Vision model"}

  User -- upload JPG -->  Convert --create child text doc --> Chunk --> ChromaDB
  User -- link to online JPG --> Fetch -- create doc --> Convert

  Convert <-- describe image --> Ollama
```

spec/services/destroy_document_spec.rb
oxaroky02: Nice. And using `let!(doc)` got rid of one of the `before` blocks. As for `let(child)` however, Rubocop doesn't like me using `let!`. 😄 

```
spec/services/reset_document_spec.rb:15:7: C: RSpec/LetSetup: Do not use let! to setup objects not referenced in tests.
      let!(:child) { create(:document, parent: doc) }
```

oxaroky02: @oxaronick, an alternative to `before` block with `child.save`, what do you think about referencing `child` in the example itself? I.e. 

```rb
    context "when document has children" do
      let(:child) { create(:document, parent: doc) }

      it "destroys both documents document" do
        child    # <<<< like so
        expect { subject.execute }.to change(Document, :count).by(-2)
      end
    end
```

oxaronick: Does `let!(:child)` not work?

oxaroky02: Using `let!(child)` works. But `ops lint` doesn't like it.

oxaronick:
Looks good in the UI!

However, on the Collection page, when I upload a new document, I see just the original doc and the tags `Created`, `Not embedded`, and `Not extracted`. The child doc card is not created, so I'm assuming any turbo stream replacements don't modify the UI as their target element is not present.

image

Also, when I delete a doc, I see the `Deleting` tag on it, but the parent doc and child doc remain on the Collection page until I refresh.

image

Do you see the same?

oxaroky02:
> However, on the Collection page, when I upload a new document, I see just the original doc and the tags `Created`, `Not embedded`, and `Not extracted`. The child doc card is not created, so I'm assuming any turbo stream replacements don't modify the UI as their target element is not present.

Yes. I have the same issue, I haven't figured out how to make the async work (jobs) cause changes to the page while ingest is happening.

mattlindsey:
> > ... describing the steps of ingesting a document and the 'philosophy' of how they're stored in Archyve. There are a lot of steps and services in the code, so it's a little hard to understand.
> 
> @mattlindsey, see PR description for some high level diagrams. These will end up in `docs/` as well. After we merge this, to ingest images and convert them into textual representation for chunking,
> 
Hi @oxaroky02.  Nice that this works.  I understand the flow/steps now, but am still a little confused about why child documents are created and will chunks always be text?  A Converter is always a TextConverter. Right? Sounds like that's the plan, but Chroma is multimodal and accepts images now directly from a URI and embedding function.  So that's an option.

And will PDF and HTML documents also have additional child documents for images and 1st level links (for example)? Can you add how that will work to the diagram.

oxaroky02:
> I understand the flow/steps now, but am still a little confused about why child documents are created and will chunks always be text? A Converter is always a TextConverter. Right? Sounds like that's the plan, but Chroma is multimodal and accepts images now directly from a URI and embedding function. So that's an option.

@mattlindsey, yeah, currently Archyve deals with text primarily and for the time being we haven't tried to change that. It certainly would be interesting to consider "image chunk" as a type of chunk; not sure how that would affect the RAG-side of the system where we would augment queries. @oxaronick and I have had arm-waving discussions about augmenting with more than text but the protocols were less able then. The Open AI and Ollama API's are evolving so fast though so maybe this is something worth considering sooner.

Issue #109, opened 2024-10-18
Model config system_prompt not being used in chat


mattlindsey:
Using the admin page, I added the system_prompt `Please call me Jeremy` to the default `phi3:latest` chat model, but when chatting I don't see it being used.  

In the api call for `Hi. Do you know my name?` I just see the message below (from Message.summary_chat method).  Shouldn't it use use or add a `role: system` message?

```
{
  "model": "phi3:latest",
  "messages": [
    {
      "role": "system",
      "content": "You are a summarization AI. You'll never answer a user's question directly, but instead summarize\nthe user's request into a single short sentence of four words or less. Always start your answer\nwith an emoji relevant to the summary\",\n"
    },
    {
      "role": "user",
      "content": "Who is the president of Gabon?"
    },
    {
      "role": "assistant",
      "content": "🇬🇦 President of Gabon"
    },
    {
      "role": "user",
      "content": "Who is Julien Chaumond?"
    },
    {
      "role": "assistant",
      "content": "🧑 Julien Chaumond"
    },
    {
      "role": "user",
      "content": "what is 1 + 1?"
    },
    {
      "role": "assistant",
      "content": "🔢 Simple math operation"
    },
    {
      "role": "user",
      "content": "What are the latest news?"
    },
    {
      "role": "assistant",
      "content": "📰 Latest news"
    },
    {
      "role": "user",
      "content": "How to make a great cheesecake?"
    },
    {
      "role": "assistant",
      "content": "🍰 Cheesecake recipe"
    },
    {
      "role": "user",
      "content": "what is your favorite movie? do a short answer."
    },
    {
      "role": "assistant",
      "content": "🎥 Favorite movie"
    },
    {
      "role": "user",
      "content": "Explain the concept of artificial intelligence in one sentence"
    },
    {
      "role": "assistant",
      "content": "🤖 AI definition"
    },
    {
      "role": "user",
      "content": "Draw a cute cat"
    },
    {
      "role": "assistant",
      "content": "🐱 Cute cat drawing"
    },
    {
      "role": "user",
      "content": "Hi. Do you know my name?"
    }
  ]
}
```

Response body:
```
🙋‍♂️ I don't have access to personal data unless it's shared with me in the context of our conversation, and even then, privacy is paramount. My primary function is to assist without retaining identifying information about individuals beyond this interaction for their safety and confidentiality.
```

nickthecook:
Yes, the system prompt is in the schema but isn't actually used.

[...]

PR #146, opened 2024-11-22
Add system_prompt support

mattlindsey:
For #109.

@nickthecook This is almost ready, but I wanted to make sure it is what is desired.  I'm adding a Message to the start of the Conversation, but maybe this should be done 'behind the scenes' on the chat call.  I like my approach, since it's more obvious what's going on, but others might think differently.

### DIFF
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -28,7 +28,6 @@ class ConversationsController < ApplicationController
     @conversation = Conversation.new
     @conversation.user = current_user
     @conversation.model_config ||= Setting.chat_model
-    @conversation.model_server ||= ModelServer.first
     @conversation.title ||= "New conversation"
     @conversation.search_collections = Setting.get(:search_collections, target: current_user)
     @conversation.save!
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -28,6 +28,7 @@ class ConversationsController < ApplicationController
     @conversation = Conversation.new
     @conversation.user = current_user
     @conversation.model_config ||= Setting.chat_model
+    @conversation.model_server ||= ModelServer.first
     @conversation.title ||= "New conversation"
     @conversation.search_collections = Setting.get(:search_collections, target: current_user)
     @conversation.save!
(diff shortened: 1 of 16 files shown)


mattlindsey:
The code in this PR runs properly (I think) but give pretty strange results with the seeded system_prompt.
I tell it to answer using 'context from the provided documents' but it doesn't know much about documents/collections/context/etc.   I think something like the XML tags in the context (in the PR we closed) is needed so it knows about document filenames, links, and what 'context' is.

nickthecook:
Will have a look!

app/controllers/conversations_controller.rb
```diff
@@ -28,6 +28,7 @@ def create
     @conversation = Conversation.new
     @conversation.user = current_user
     @conversation.model_config ||= Setting.chat_model
+    @conversation.model_server ||= ModelServer.first
```
oxaronick, line 31: You can use `ModelServer.active_server` for this.

oxaronick: Actually, why does Conversation have a ModelServer?

oxaronick:
Hopefully using `message.author = nil` to denote a system message will make some of the problems with the API and ApiCall go away.

I tried out a couple of system calls using another chat app that used the Archyve API for chat. It was hard - I didn't come up with a prompt that really gives the sense of an archivist or something else in charge of a body of data, and just mediating the user's interactions with the data. Hopefully you do better than I did!

mattlindsey:
I think this PR works now and all tests pass, but the title of every new conversation seems to stay 'New conversation.'  So something may have broken.

And I'm sure you'll want to see if the prompt helps.  Which chat app are you using the Archyve API from?  I should try that too.

oxaronick:
I'm also seeing that Conversations are not being summarized. Looks like the system message has broken this rather poor line of code in MessagesController:

```ruby
TitleSetterJob.perform_async(@conversation.id) if @conversation.messages.count == 1
```

That should probably be a helper method or something, and should now check if this is the first _user_ message in the conversation.

mattlindsey:
I think everything works now and this PR may be a good first step.

If you want, we can just make the seeded prompt this: `You are Archyve, a helpful AI assistant.`  Then leave it up to a later PR to try to make it be more helpful.  Or what is there now is probably fine.

A nice next step might be:
- Display and allow editing of global system_prompt on /settings page

nickthecook:
This looks good.

That would be a good next step!

PR #155, opened 2024-12-06
Add conversion-related states to Document

oxaroky02:
A small one this time. 😄 

- Add `converting` and `converted` states to `Document`
- Add `converting` and `convert` events
- Ensure conversion job updates state before and after conversion
- Also ... 
  - Hacked `after_update_commit` to call `parent.broadcast_replace if parent` which gets rid of the bug where occasionally the collection view doesn't update. This has the side effect that if a document has multiple children they will cause the entire parent to update. 

> Destroying a document still has inconsistant view update.


oxaronick:
Sorry for letting this one sit!

